PR #6, opened 2014-09-16
Add impersonation method

dentarg:
Support for https://docs.auth0.com/auth-api#impersonation.


dentarg:
Sorry for the delay, I had to prioritize other things. I will await the merge of #9 before continuing here.

PR #7, opened 2014-09-16
Add create user method

dentarg:
Support for [creating a user](https://docs.auth0.com/api#!#post--api-users).

Will raise exception and echo the error message from the API if the call wasn't successful.


dentarg:
Sorry for the delay, I had to prioritize other things. I close this as it's covered by #9.

PR #10, opened 2015-03-10
updated webmock version, updated documentation

offtop:


jfromaniello:
excuse my ignorance, what do you use to generate the docs? Thanks a lot

dentarg:
Do the docs need to be part of the repo?

offtop:
I used http://yardoc.org/ (just run yard in repo)
@dentearg I don;t really like them here as well, but it seems to be the best way. Any Ideas ?

dentarg:
I think the user can generate the docs or use http://www.rubydoc.info/github/auth0/ruby-auth0

dentarg:
Created #14

PR #12, opened 2015-03-11
Fix typo delegation signature; open_id => openid

dentarg:
According to https://auth0.com/docs/protocols it's "openid".


dentarg:
I have not tested this against the API.

jfromaniello:
Yes, it looks correct. Thank you!

PR #13, opened 2015-03-11
Make the client work for API v1

dentarg:
We should probably have some integration tests, right? :)

### DIFF
--- a/lib/auth0/mixins/initializer.rb
+++ b/lib/auth0/mixins/initializer.rb
@@ -21,7 +21,7 @@ module Auth0
           @client_secret  = options[:client_secret]
           @token          = obtain_access_token
         end
-        self.class.headers "Authorization" => "Bearer #{options[:access_token]}"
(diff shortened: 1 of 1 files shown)


offtop:
Critical update, please merge it.

jfromaniello:
Thanks, released as 3.1.2.

Regarding integration tests.. what can we do?

dentarg:
Dedicate an auth0 account for tests, keys can be encrypted in .travis.yml, write tests that actually do stuff with the API. What do you think?

dentarg:
Created #15

dentarg:
> Thanks, released as 3.1.2.

@jfromaniello did you forget to push?

https://github.com/auth0/ruby-auth0/blob/3d2385890b619f1287bdfc33abc348b9e7cd4d51/lib/auth0/version.rb#L3 still says 3.1.1

dentarg:
ping @jfromaniello

PR #18, opened 2015-03-24
fix token initialization to prevent nil token value

ethancrawford:


jfromaniello:
Thanks I will push the gem soon

PR #19, opened 2015-03-24
Update readme.md api v2 basic usage example

ethancrawford:
By following the existing api v2 basic usage example, a rather obscure error is generated - 'SocketError: getaddrinfo: Name or service not known'. Including this key/value pair as in api v1 makes it work.


jfromaniello:
Thanks man!

PR #21, opened 2015-03-25
make AuthenticationEndpoints available to client in v2 mode

MungoH:
This is a fix for https://github.com/auth0/ruby-auth0/issues/20
It implies :client_id is required when creating a v2 client.

PR #22, opened 2015-04-05
Fixed few issues from issues list

offtop:
https://github.com/auth0/ruby-auth0/issues/20
https://github.com/auth0/ruby-auth0/issues/17
https://github.com/auth0/ruby-auth0/issues/14

PR #24, opened 2015-04-14
Authendpoint in v2

Maxim-Filimonov:
Exact copy of https://github.com/auth0/ruby-auth0/pull/21 except this is from a separate branch. Please, merge one of them as it's blocking usage of v2 api


hzalaz:
@Maxim-Filimonov thanks for the PR, we will close this because we merged #22 that had a couple more issues solved

PR #27, opened 2015-04-15
Chore integration tests

hzalaz:


dentarg:
@hzalaz would be sweet if we could to like every other ruby project and use two spaces for tabs :)

hzalaz:
:+1:

hzalaz:
@dentarg spacing should be correct by now, send me a PR if you find another file with 4 spacing

PR #28, opened 2015-04-16
Fix broken link in README

dentarg:


hzalaz:
Seems travis has a restriction with external PR and secure environment variables (which kinda makes sense), so we'll have to run the integration tests conditionally :disappointed:

PR #30, opened 2015-05-05
allow setting impersonation ttl

seamusabshere:
defaults to 120 seconds per @woloski


hzalaz:
thanks @seamusabshere for the PR. Will release a new version during the day.

hzalaz:
@seamusabshere ttl for impersonation is available in the latest release.

PR #31, opened 2015-05-07
Client information headers

benschwarz:
I'd like to discuss the idea of adding some client information to our API libraries — these changes would need to be implemented in a all of our official clients, as well as requiring some effort on the devops, and data tracking sides. I'll do my best to explain why this could be such a valuable enhancement.

---

### What I want to do

Add two headers to all of our API clients when they're talking to auth0 APIs. These headers would carry the "platform" of the client, and the "version", so in the case of this Ruby client, it'd have the following two headers: 

```
User-Agent: Ruby/2.2.1 (Rails 4.2.2)
Auth0-Client: Ruby/3.4.1
```

To break it down a little bit: The client is "ruby" (language/sdk), and the version of ruby that we're using is `2.2.1`. The client version corresponds to the semantic version used for this gem (at next bump, it'll be `3.4.1`)

In addition, the each client library should allow "overloading" the client, for example, this ruby library _might_ be used in rails applications. For the `omniauth` library, we could set `Auth0::ClientVersion` to `Ruby/2.2.1 (Rails 4.2.2)`.

Client side example: 
* `User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/42.0.2311.135 Safari/537.36`
* As a querystring: `?auth0_client=auth0.js/1.1.1`

---

##### Support & CSE
I understand I'm speaking _for_ team @eugeniop a little here, but I can foresee that adding client information to an given companies requests (then store them in something like intercom) would make for really useful information when trying to help diagnose client issues, or help us point to relevant documentation. 

##### Developer advocacy
Understanding the volume and capacity of clients that are in use (and also tracking package downloads from rubygems, npm, bower etc) helps the advocacy team understand the effect/results of efforts with vertical conferences and events (Eg, if we speak at a ruby event, do we get a client bump? Do we get sign ups?). Additionally, we can work harder to improve documentation or SDKs for the more popular platforms (or, perhaps the UNpopular ones). In short, data helps us better understand our work. 

##### Statistics
screen shot 2015-05-07 at 12 34 57 pm

(Mocked graph) — We could show API requests by client type, and then also intersect that information by region, or customer details.

Further to this, we could identify customers that are using older versions of clients that require updates, or help them deal with deprecations. 

Webtask to store this in Google analytics anyone? (cc @tjanczuk )

---

### Effort required to make it a reality

- [ ] Server side logging needs to be able to log API headers somewhere (Redshift, Librato, or Google Analytics would be an obvious place)
- [ ] Update PHP client
- [ ] Update auth0.js
- [ ] Update iOS Lock
- [ ] Update Android Lock
- [ ] Update Auth0 Xamarin Client
- [ ] Update clients that I missed above

The changes to the Ruby client were conducted in less than an hour, and while this isn't an exact guide for all clients, its a good indication that we should be able to knock through the PRs pretty easily. 

---

### Feedback
Clearly, I'm not yet familiar with pitching features at auth0, nor am I aware of the entire way that this stack works. If you've got thoughts, pointers or general hand holding advice, please leave it below in the comments. From here, I can do my best to drive this through to the end.

### DIFF
--- a/spec/integration/lib/auth0/auth0_client_spec.rb
+++ b/spec/integration/lib/auth0/auth0_client_spec.rb
@@ -63,7 +63,7 @@ describe Auth0::Client do
     end
 
     it "sets the client version" do
-      headers['X-Auth0-Client'].should eql Auth0::VERSION
+      headers['X-Auth0-Client'].should eql "Ruby/#{Auth0::VERSION}"
     end
   end
 end
--- a/spec/integration/lib/auth0/auth0_client_spec.rb
+++ b/spec/integration/lib/auth0/auth0_client_spec.rb
@@ -47,23 +47,23 @@ describe Auth0::Client do
     let(:headers) { client.class.headers }
 
     it "has the correct headers present" do
-      headers.keys.sort.should eql ['Authorization', 'Content-Type', 'User-Agent', 'X-Auth0-Client']
+      expect(headers.keys.sort).to eql ['Authorization', 'Content-Type', 'User-Agent', 'X-Auth0-Client']
     end
 
     it "uses the correct access token" do
-      headers['Authorization'].should eql "Bearer abc123"
+      expect(headers['Authorization']).to eql "Bearer abc123"
     end
 
     it "is always json" do
-      headers['Content-Type'] = 'application/json'
+      expect(headers['Content-Type']).to eql 'application/json'
     end
 
     it "sets the ruby version" do
-      headers['User-Agent'].should eql "Ruby/#{RUBY_VERSION}"
+      expect(headers['User-Agent']).to eql "Ruby/#{RUBY_VERSION}"
     end
 
     it "sets the client version" do
-      headers['X-Auth0-Client'].should eql "Ruby/#{Auth0::VERSION}"
+      expect(headers['X-Auth0-Client']).to eql "Ruby/#{Auth0::VERSION}"
     end
   end
 end
--- a/spec/integration/lib/auth0/auth0_client_spec.rb
+++ b/spec/integration/lib/auth0/auth0_client_spec.rb
@@ -42,4 +42,28 @@ describe Auth0::Client do
     let(:credentials) { v2_credentials.merge({access_token: ENV["MASTER_JWT"]}) }
   end
 
+  context "client headers" do
+    let(:client) { Auth0::Client.new(v2_credentials.merge({access_token: 'abc123', domain: 'myhost.auth0.com'})) }
+    let(:headers) { client.class.headers }
+
+    it "has the correct headers present" do
+      headers.keys.sort.should eql ['Authorization', 'Content-Type', 'X-Auth0-Client', 'X-Auth0-Client-Version']
+    end
+
+    it "uses the correct access token" do
+      headers['Authorization'].should eql "Bearer abc123"
+    end
+
+    it "is always json" do
+      headers['Content-Type'] = 'application/json'
+    end
+
+    it "sets the ruby version" do
+      headers['X-Auth0-Client'].should eql "Ruby/#{RUBY_VERSION}"
+    end
+
+    it "sets the client version" do
+      headers['X-Auth0-Client-Version'].should eql Auth0::VERSION
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


woloski:
+1000

woloski:
@ntotten had this on the roadmap I think, but the sooner we do it the better

mgonto:
Nice Ben!! It was on our backlog as well but it's nice to see it implemented in the first SDK! 
Good job! 




On Thu, May 7, 2015 at 3:50 PM -0700, "Matias Woloski" <notifications@github.com> wrote:












@ntotten had this on the roadmap I think, but the sooner we do it the better

eugeniop:
Yes, @ntotten, @benschwarz and I are on the same room here :-)

woloski:
:8ball:

jfromaniello:
:+1: good idea

hzalaz:
@benschwarz great, Lucky I only have to add the `X-Auth0-Client` header to iOS/Android :smile:

spec/integration/lib/auth0/auth0_client_spec.rb
```diff
@@ -42,4 +42,28 @@
     let(:credentials) { v2_credentials.merge({access_token: ENV["MASTER_JWT"]}) }
   end
 
+  context "client headers" do
+    let(:client) { Auth0::Client.new(v2_credentials.merge({access_token: 'abc123', domain: 'myhost.auth0.com'})) }
+    let(:headers) { client.class.headers }
+
+    it "has the correct headers present" do
+      headers.keys.sort.should eql ['Authorization', 'Content-Type', 'User-Agent', 'X-Auth0-Client']
```
hzalaz: I'd rather use _expect_ syntax instead of _should_.

spec/integration/lib/auth0/auth0_client_spec.rb
```diff
@@ -42,4 +42,28 @@
...
+    end
+
+    it "uses the correct access token" do
+      headers['Authorization'].should eql "Bearer abc123"
+    end
+
+    it "is always json" do
+      headers['Content-Type'] = 'application/json'
```
hzalaz: I think you forgot the `eq` :smile:

spec/integration/lib/auth0/auth0_client_spec.rb
hzalaz: Unless we use implicit subject..

spec/integration/lib/auth0/auth0_client_spec.rb
```diff
@@ -42,4 +42,28 @@
     let(:credentials) { v2_credentials.merge({access_token: ENV["MASTER_JWT"]}) }
   end
 
+  context "client headers" do
```
hzalaz: I think it will better to test this in an unit tests but we can have this here till we re-do them

ntotten:
I just read the spec on the use of `X-` headers. It seems the new recommendation is to just prefix with organization name so we might as well just use `Auth0-Client` without the X-.

hzalaz:
That's true @ntotten, also we can add all the information in the `User-Agent` but either way is fine

ntotten:
The one thing about user agent is that it doesnt work in the case of lock where we would add the version to the /authorize query string. `/authorize?auth0_client=lock%2F3.4.5`

hzalaz:
@ntotten :+1: you're right

spec/integration/lib/auth0/auth0_client_spec.rb
benschwarz: Seriously funny :smile: 
… I think I did this when I was hanging with @eugeniop and @ntotten. I'll amend :-)

benschwarz:
I've updated the proposal to not use `X-Auth0-Client`, but `Auth0-Client` instead. 
Thanks for the feedback on that folks :leaves: 

@hzalaz I updated to use the newer `expect` syntax also. 

Finally, the build on this project seems to be a bit flakey — fails then pretty much always passes on second run.

hzalaz:
@benschwarz yes, I've thought the brittle build was that fixed but I must have missed something. Will merge it now and fix the build later.

PR #32, opened 2015-05-29
Use client header spec as per documentation

benschwarz:
Update client version headers as per specification


hzalaz:
@benschwarz beat me to it (glad cause its way down in my TODO list). Just rebase it against master for the build to be fixed

PR #42, opened 2016-01-05
Adding support for v2 password change ticket endpoint

LindseyB:
This adds support for the v2 password change ticket creation endpoint.


ntotten:
This is already in v2 which we are merging in a few min.

PR #45, opened 2016-01-08
Support ruby 1.9.3

singhals:


ntotten:
Thanks for the issue. We have decided that given the age of Ruby 1.9.3 that we will not support any version less than 2.0.0. See here for more context: http://blog.sass-lang.com/posts/560719

Please feel free to comment if you disagree. We are happy to discuss, but we have seen almost no interest in this and I suspect that the interest will continue to drop over time.

Issue #63, opened 2016-04-28
Exception messages are ruby hash string literals


chazu:
when an API error results in a BadRequest exception being thrown, client code using the gem only has access to the exception message as a ruby hash literal - meaning that eval must be called to change it into a format from which the actual message or error code can be extracted.

While regexp matching could be used as a slightly-more-secure workaround, in cases where either exception error messages could be forwarded to end users or other logic is in place which examines the response body when an exception is thrown, this makes it clunky to work with exception data. If the request body were JSON-encoded before being set to the message of an exception, client code could at least parse the exception's message into a hash when thrown.

PR #66, opened 2016-05-09
put json in exceptions as messages, so they can be parsed without #eval

chazu:
Per issue #63 , exceptions thrown after http requests will contain json strings rather than strings of ruby hash literals, allowing manipulation by client code without using eval.


ianks:
Heads up, I just made a PR which does the same thing :fearful: 

 Mine is slightly better since it has tests and removes a redundant JSON parse and generation.

https://github.com/auth0/ruby-auth0/pull/67

chazu:
Roger that, i'll close this puppy. Thanks.

PR #67, opened 2016-05-12
Use JSON (result.body) for exception messages in requests

ianks:
Previously, Auth0::Exception was being fed a Hash for its message content.
All messages are stringified, so when a user accesses the error message,
they are left with a string representation of a hash. Here is an example.

```ruby
def some_method
  auth0_client.user('notarealuser')
rescue Auth0::Exception => e
  puts e.message # => "{\"statusCode\"=>400, \"error\"=>\"Bad Request\", \"message\"=>\"Path validation error: 'String does not match pattern ^.+\\\\|.+$: 3241312' on property id (The user_id of the user to retrieve).\", \"errorCode\"=>\"invalid_uri\"}"

  # cannot parse this error, :(
end
```

These hash representations do not have an efficient, robust parsing mechanism.
Therefore, I propose we simply use the JSON representation from the response
body as the message, so users can parse the message and do what they need
with it. A new example would work like this:

```ruby
def some_method
  auth0_client.user('notarealuser')
rescue Auth0::Exception => e
  JSON.parse(e.message) #=> {"statusCode"=>400, "error"=>"Bad Request", "message"=>"Path validation error: 'String does not match pattern ^.+\\|.+$: 3241312' on property id (The user_id of the user to retrieve).", "errorCode"=>"invalid_uri"}

  # yay!
end
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,11 @@
+---
 language: ruby
 sudo: false
 cache: bundler
 rvm:
-- 2.1.0
-- 2.2.1
+  - 2.1.9
+  - 2.2.5
+  - 2.3.1
 script: "./build_travis.sh"
 after_success: "./deploy_documentation.sh"
 deploy:
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ source 'http://rubygems.org'
 gemspec
 
 group :development do
-  gem 'terminal-notifier-guard', require: false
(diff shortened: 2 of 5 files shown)


ianks:
@auth0 I fixed the travis build as well. `listen` is not compatible with versions of ruby less than 2.2.3, so I prevented listen from being installed on travis.

PR #68, opened 2016-06-07
Fix start_passwordless_email_flow

mtmorra:
Updated start_passwordless_email_flow endpoint to match the documentation on auth0.com.


chenkie:
Addressed with recently merged 4.1.0 PR.

PR #70, opened 2016-07-13
Obtain user tokens

fleko:
Adds a new method to the authentication endpoints to retrieve the user access and id tokens


fleko:
Created in error

PR #73, opened 2016-09-29
Typo

amingilani:


amingilani:
The current code for the API is incorrect, and not specifying `:api_version => "2"` defaults to `:api_version => "1"` with the error `Auth0::InvalidCredentials: Invalid API v1 client_id and client_secret`

You should've hired me, guys! 😁

amingilani:
This code was still incorrect, sorry. It appears as thought he `:api_version` cannot be a string and has to be an integer.

amingilani:
All fixed now.

blueboxjesse:
Similar to #78. Suggest reviewing and merging both simultaneously.

PR #80, opened 2017-01-11
Remove use of %i shorthand

anthonylebrun:
Older ruby versions do not support this syntax


anthonylebrun:
Whoops, didn't mean to do that.

PR #104, opened 2018-02-01
Add scripts and Dockerfile to publish SDK on rubygems

alexisluque:


.dockerignore
```diff
@@ -0,0 +1,6 @@
+bin
```
lbalmaceda: All these lines refer to files not pushed to the repo. Remember that .dockerignore already ignores the files and folders listed under the .gitignore file.

exec.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+docker build -t auth0-publish-rubygem .
+docker run -it auth0-publish-rubygem
```
lbalmaceda: Will the container interactive shell close after running the "default command"? and in that case, shouldn't you add the `--rm` flag to also shutdown the container?

RUBYGEM.md
```diff
@@ -0,0 +1,9 @@
+# Publish the Gem on RubyGems.org
+
+To publish the gem set `RUBYGEMS_EMAIL` and `RUBYGEMS_PASSWORD` environment variables with your email and password from your RubyGems account respectively.
+Then run the instructions to build and run the docker image to build and publish the gem.
```
lbalmaceda: "Then run the following [Docker](https://docs.docker.com/engine/installation/) commands in the terminal to build and publish the gem."

PR #105, opened 2018-02-12
Add support to /api/v2/users-by-email

edgurgel:
Hey, we would like to use the `users-by-email` API. I've tested locally against auth0 and it seems to be working fine. Should I also add an integration test?

Thanks!

API docs https://auth0.com/docs/api/management/v2#!/Users_By_Email/get_users_by_email


edgurgel:
Thanks! 🎉

PR #108, opened 2018-02-23
Add ability to login via the oauth/ro endpoint

psparrow:
This PR introduces `AuthenticationEndpoints.login_with_legacy_endpoint` to post username and password credentials to the "/oauth/ro" endpoint that is currently still available for some auth0 accounts. 

This makes it possible to reap the benefits of upgrading the auth0 gem including upgrading to rest-client 2.x without having to change the `grant_type` or introduce `audience` and `realm` parameters that the `.login` method does not yet support.

### DIFF
--- a/lib/auth0/api/authentication_endpoints.rb
+++ b/lib/auth0/api/authentication_endpoints.rb
@@ -70,6 +70,26 @@ module Auth0
         post('/oauth/token', request_params)
       end
 
+      # Login using username/password with the /oauth/ro endpoint
+      # @see https://auth0.com/docs/auth-api#!#post--oauth-ro
+      # @param username [string] Username
+      # @param password [string] User's password
+      # Active Directory/LDAP, Windows Azure AD and ADF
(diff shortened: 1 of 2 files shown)


aaguiarz:
Just to understand better, what is blocking you from using the existing endpoint that uses '/oauth/token'?

psparrow:
Posting to oauth/token with the same credentials we currently use to post to oauth/ro returns an RS256 token instead of an HS256 token. So, to start using the new endpoint, we either need to update our auth0 account settings or modify the code that uses the token. Until we can do that, using the oauth/ro endpoint is our best option. Plus, it's still supported for other login methods, like `.phone_login`.

joshcanhelp:
@psparrow - I understand your use case here but I'm a little hesitant to add a method that is getting deprecated in the other SDKs we maintain. We're discussing internally now 👍 either way, appreciate the contribution!

psparrow:
No problem. I think we’ve already switched to the newer endpoint anyway.

Issue #109, opened 2018-03-27
Incorrect request_params for username-password login


mmaddex:
This is preventing callers from passing `audience` and `realm` along.

There seem to be a few things that could be addressed here:

- This should include `audience` and `realm`
- Remove old `/oauth/ro` params: `connection`, `id_token`, and `device`
- the default `grant_type` should be `'password'` not `password`

https://github.com/auth0/ruby-auth0/blob/687326f7c81db114a1ff8de2047c8974fa461247/lib/auth0/api/authentication_endpoints.rb#L59-L70

gastonrey:
Am having problems to login via user/password due to the deprecated 'Connection' key, it's now `realm`, so I have changed the login method to:

```
request_params = { 
   client_id:     @client_id, 
   client_secret: @client_secret, 
   username:      username, 
   password:      password, 
   scope:            options.fetch(:scope, 'openid'), 
   realm:            connection_name, 
   grant_type:    options.fetch(:grant_type, 'password'), 
   id_token:      id_token, 
   device:        options.fetch(:device, nil) 
 } 
```
Replaced `connection` key by `realm`.

And then at my login request the `grant-type` is supposed to be passed as:

```
grant_type: 'http://auth0.com/oauth/grant-type/password-realm'
```
Just inside the `options` object. Would probably be a good idea to leave it as the default value when user/password login.

joshcanhelp:
Apologies for the late reply here ... the auth endpoints module in this SDK is getting an overhaul for the next release. We'll be deprecating (not removing yet) a few methods, this one included, and replacing with ones that more closely match how the endpoints work currently. All of the concerns here will be addressed. 

Thank you for the report!

j-collier:
> We'll be deprecating (not removing yet) a few methods, this one included, and replacing with ones that more closely match how the endpoints work currently. All of the concerns here will be addressed.

Are the plans for this public? I'd love to help tackle this personally.

joshcanhelp:
@j-collier - I appreciate the offer! The plans are not public but only because our tracking is done in the same system as our product. Happy to share what we have in mind remaining.

Much of this is complete, some of which has already been merged:

- #129 (merged) adds a formal client credentials grant
- #130 (merged) adds a `/userinfo` call that works (have to include an access token)
- #131 (in review) will add a method to perform an authorization code exchange
- #133 (in review) will add a method that does a more complete resource owner grant

The remaining methods that have not been started:

[...]

PR #107, opened 2018-02-22
Correctly default grant_type to 'password' in login method

psparrow:
This PR changes the default `grant_type` option in `AuthenticationEndpoints#login` from the value of the password parameter to "password". It also updates the spec to use a test password other than "password" to verify that the default grant type is set correctly.

PR #110, opened 2018-03-28
Support optional params for oauth/token endpoint for username-password login

leomao10:
This PR will fix issue #109 by adding support to pass `realm` and `audience` and also include fixes for #107 

I still keep some of the outdated params like `connection` for backward compatibility, but I think it should be removed eventually

### DIFF
--- a/lib/auth0/api/authentication_endpoints.rb
+++ b/lib/auth0/api/authentication_endpoints.rb
@@ -45,12 +45,14 @@ module Auth0
       end
 
       # Logins using username/password
-      # @see https://auth0.com/docs/auth-api#!#post--oauth-ro
+      # @see https://auth0.com/docs/api/authentication#resource-owner-password
       # @param username [string] Username
       # @param password [string] User's password
-      # @param scope [string] Defaults to openid. Can be 'openid name email', 'openid offline_access'
-      # @param id_token [string] Token's id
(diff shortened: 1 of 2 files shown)


joshcanhelp:
Thanks for the contribution here. I'm going to merge #107 first and then ask you to rebase against that, just FYI.

mmaddex:
I'm curious what backward compatibility keeping `connection`, `id_token`, and `device` provides, given that this method is now hitting an endpoint that doesn't use those params anyway?

leomao10:
Yeah, I guess it is a breaking change anyway? happy to remove those params if you want

joshcanhelp:
@mmaddex - Let's do that in a separate PR. Can't do much about the `id_token` parameter but we can at least deprecate it and remove from the request. 

@leomao10 - For now, let's not expand the scope on this particular PR.

joshcanhelp:
@leomao10 - are you able to rebase and resolve the conflicts here? Also a couple non-blocking questions for you above. 

thank you!

leomao10:
Yup, will do it tonight when I have time, cheers

joshcanhelp:
@leomao10 - Last checkin on this. Are you able to rebase and get this completed? Let me know if not, I can get it into shape in a branch here. Thanks!

joshcanhelp:
We're working on getting the auth endpoints in better shape and will be deprecating this method to clean it up a bit.

PR #113, opened 2018-06-20
Add pagination and tests to Clients and Connections endpoints

joshcanhelp:
- Add new `:page` and `:per_page` method parameters to `Auth0::Api::V2::Clients` and `Auth0::Api::V2::Connections` modules.
- Add ability to pass an `Array` for the `fields:` parameter and convert to comma-sep string
- Add tests for all changes above.

### DIFF
--- a/spec/lib/auth0/api/v2/clients_spec.rb
+++ b/spec/lib/auth0/api/v2/clients_spec.rb
@@ -5,18 +5,62 @@ describe Auth0::Api::V2::Clients do
     dummy_instance.extend(Auth0::Api::V2::Clients)
     @instance = dummy_instance
   end
+  
   context '.clients' do
     it { expect(@instance).to respond_to(:clients) }
     it { expect(@instance).to respond_to(:get_clients) }
+
     it 'is expected to send get request to /api/v2/clients/' do
-      expect(@instance).to receive(:get).with('/api/v2/clients', fields: nil, include_fields: nil)
+      expect(@instance).to receive(:get).with(
+        '/api/v2/clients',
+        fields: nil,
+        include_fields: nil,
+        page: nil,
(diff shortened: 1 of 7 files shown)


spec/lib/auth0/api/v2/clients_spec.rb
```diff
@@ -5,18 +5,62 @@
...
+        fields: 'name,app_type',
+        page: nil,
+        per_page: nil
+      )
+      expect {
+        @instance.clients(fields: ['name','app_type'], include_fields: true)
+      }.not_to raise_error
+    end
+
+    it 'is expected to add pagination' do
```
machuga: This is asserting something about adding pagination, but only tests that an error wasn't raised. Is there an indicator that the params weren't silently ignored?

joshcanhelp: So ... that first `expect` line sets up the second one. If you change the `page` value in `with()` to a different value, it fails.

machuga:
Right, but I'm suggesting that code wise I cannot tell that what is being
tested is ensuring pagination works.

joshcanhelp:
@machuga - As in, you can't tell if the results themselves are paginated? I can add an integration test for that. 

But, for a unit test, I *think* this works ok. It was a bit of trial and error to get it to work but changing the values of, say, the `page` parameter on one or the other but not both would make the test fail. So that first `expect` is saying "make the instance get an endpoint with the following parameters" and the second calls that instance, causing an error if it does not meet the expectation. [Source](https://relishapp.com/rspec/rspec-mocks/docs/basics/expecting-messages).

machuga:
Ya I think the test name just implies that something else is being tested. `it ......`

PR #115, opened 2018-06-28
Add search_engine parameter to Users endpoint + tests

joshcanhelp:
- Replace the hard-coded search engine version `v2` with a modifiable one
- Improve docs for the method
- Add unit and integration tests
- Remove skipped user test

### DIFF
--- a/lib/auth0/api/v2/users.rb
+++ b/lib/auth0/api/v2/users.rb
@@ -5,18 +5,21 @@ module Auth0
       module Users
         attr_reader :users_path
 
-        # Retrieves a list of existing users.
-        # @see https://auth0.com/docs/api/v2#!/Users/get_users
-        # @param per_page [integer] The amount of entries per page. Default: 50. Max value: 100.
-        # @param page [integer]  The page number. Zero based.
-        # @param include_totals [boolean] True if a query summary must be included in the result.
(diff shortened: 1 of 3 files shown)


lib/auth0/api/v2/users.rb
```diff
@@ -27,9 +30,9 @@ def users(options = {})
             connection:     options.fetch(:connection, nil),
             fields:         options.fetch(:fields, nil),
             include_fields: options.fetch(:include_fields, nil),
-            q:              options.fetch(:q, nil)
+            q:              options.fetch(:q, nil),
```
machuga: Nothing you need to worry about now since the convention is used, but just using as an opportunity to share info:

`.fetch` is typically used on a hash when you want to look for a key and raise an exception if it is not empty, or return a specific default. Accessing a hash key via the `[]` method is equivalent to `.fetch(key, :nil)`, so when `nil` is the desired default return value, you'll typically find the `[]` method is used.

joshcanhelp: @machuga - So something like:

```ruby
request_params = {
  ...
  q: [:q]
  ...
}
```

?

machuga: Close: 
```ruby
options[:key] === options.fetch(:key, nil)
```

PR #116, opened 2018-07-05
Add pagination and tests for Client Grants

joshcanhelp:
- Add `page` and `per_page` parameters to `Auth0::Api::V2::ClientGrants::client_grants` endpoint
- Add unit and integration tests for this change
- Remove `sleep` directives from the client grant integration tests

### DIFF
--- a/lib/auth0/api/v2/client_grants.rb
+++ b/lib/auth0/api/v2/client_grants.rb
@@ -7,10 +7,15 @@ module Auth0
 
         # Retrieves a list of all client grants.
         # @see https://auth0.com/docs/api/management/v2#!/client_grants/get_client_grants
-        #
+        # @param page [int] Page number to get, 0-based.
+        # @param per_page [int] Results per page if also passing a page number.
         # @return [json] Returns the client grants.
-        def client_grants
-          get(client_grants_path)
+        def client_grants (page: nil, per_page: nil)
+          request_params = {
+            page: !page.nil? ? page.to_i : nil,
+            per_page: !page.nil? && !per_page.nil? ? per_page.to_i : nil
+          }
+          get(client_grants_path, request_params)
         end
         alias get_all_client_grants client_grants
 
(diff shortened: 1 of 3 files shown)


lib/auth0/api/v2/client_grants.rb
```diff
@@ -7,10 +7,15 @@ module ClientGrants
...
-        #
+        # @param page [int] Page number to get, 0-based.
+        # @param per_page [int] Results per page if also passing a page number.
         # @return [json] Returns the client grants.
-        def client_grants
-          get(client_grants_path)
+        def client_grants (page: nil, per_page: nil)
+          request_params = {
+            page: !page.nil? ? page.to_i : nil,
+            per_page: !page.nil? && !per_page.nil? ? per_page.to_i : nil
```
machuga: You may want to simplify this down to `page.nil? || per_page.nil? ? nil : per_page.to_i`. There are a lot of symbols going on already, so reducing the cognitive overhead of negated `AND`s should help reduce the complexity.

I'll also add that if anyone drops in a boolean to these params, the method will raise since `to_i` is not implemented on booleans. I'm assuming this logic exists just to cast strings to ints though. And some of the logic seems confusing to read through. Here is a potential alternative to the method that I'm not sure I like any better, but 🤷‍♂️.

```ruby
def client_grants(page: nil, per_page: nil)
  return get(client_grants_path) if page.nil?

  get(client_grants_path, page: page.to_i, per_page.nil? ? nil : per_page.to_i) 
end
```
It explicitly shows that if `page` is `nil` we have nothing to do here, then continues on with the logic.

Ideally you could get rid of the last ternary if the logic of passing `per_page` supports it, but that's something you'd have to check out.

joshcanhelp: Good feedback, thank you. I'll see if I can simplify this a little better in a new commit.

PR #117, opened 2018-07-05
Add pagination and tests for Rules endpoint

joshcanhelp:
- Add `page` and `per_page` parameters to `Auth0::Api::V2::Rules::rules` endpoint
- Add unit and integration tests for this change
- Remove `sleep` directives from the rules integration tests

### DIFF
--- a/lib/auth0/api/v2/rules.rb
+++ b/lib/auth0/api/v2/rules.rb
@@ -14,14 +14,18 @@ module Auth0
         # @param fields [string] A comma separated list of fields to include or exclude from the result.
         # @param include_fields [boolean] True if the fields specified are to be included in the result, false otherwise.
         # @param stage [string] Retrieves rules that match the execution stage (defaults to login_success).
+        # @param page [int] Page number to get, 0-based.
+        # @param per_page [int] Results per page if also passing a page number.
         #
         # @return [json] Returns the existing rules.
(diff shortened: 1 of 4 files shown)


lib/auth0/api/v2/rules.rb
```diff
@@ -14,14 +14,18 @@ module Rules
...
-        def rules(enabled: nil, fields: nil, include_fields: nil, stage: nil)
+        def rules(enabled: nil, fields: nil, include_fields: nil, stage: nil, page: nil, per_page: nil)
           request_params = {
             enabled:          enabled,
             fields:           fields,
             include_fields:   include_fields,
-            stage:            stage
+            stage:            stage,
+            page:             !page.nil? ? page.to_i : nil,
+            per_page:         !page.nil? && !per_page.nil? ? per_page.to_i : nil
```
machuga: Going to give this the same feedback as the other one where this logic is confusing to read through. I'm noticing the other params aren't being cast in any way. Can you elaborate more on why we are casting these vars?

joshcanhelp: Since I'm adding new params here, I was only casting the additions, not the others.

joshcanhelp:
@machuga - I should have done this before but ... I tested the params as strings and got the correct results. I'm not seeing any reason to cast these at all. Pushed that change (plus adding back a few of the `sleep` directives since I was getting a 429). 

I'll make the same changes in #116

PR #120, opened 2018-07-17
Clean up auth endpoint tests

joshcanhelp:
- Change `URI.escape` -> `CGI.escape`, as recommended by Rubocop
- Formatting and cleanup on tests for methods that are not being replaced/deprecated


cocojoe:
@joshcanhelp do these tests still un locally ok?

joshcanhelp:
@cocojoe - Yes, mostly whitespace changes

PR #123, opened 2018-07-24
Add additional parameters for WS-Fed URL

joshcanhelp:
Add `options` parameter to `wsfed_url` method + tests

### DIFF
--- a/lib/auth0/api/authentication_endpoints.rb
+++ b/lib/auth0/api/authentication_endpoints.rb
@@ -215,12 +215,22 @@ module Auth0
       # Return a WS-Federation URL.
       # @see https://auth0.com/docs/api/authentication#accept-request35
       # @param connection [string] Connection to use; empty to show all
+      # @param options [hash] Extra options; supports wtrealm, wctx, wreply
       # @return [url] WS-Federation URL
-      def wsfed_url(connection = UP_AUTH)
+      def wsfed_url(connection = UP_AUTH, options = {})
         request_params = {
-          whr: connection
+          whr: connection,
+          wtrealm: options[:wtrealm],
+          wctx: options[:wctx],
+          wreply: options[:wreply]
         }
-        URI::HTTPS.build(host: @domain, path: "/wsfed/#{@client_id}", query: to_query(request_params))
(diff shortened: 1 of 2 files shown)


lib/auth0/api/authentication_endpoints.rb
```diff
@@ -215,12 +215,22 @@ def samlp_url(connection = UP_AUTH)
...
         request_params = {
-          whr: connection
+          whr: connection,
+          wtrealm: options[:wtrealm],
+          wctx: options[:wctx],
+          wreply: options[:wreply]
         }
-        URI::HTTPS.build(host: @domain, path: "/wsfed/#{@client_id}", query: to_query(request_params))
+
+        url_client_id = @client_id if !request_params[:wtrealm]
```
lbalmaceda: so `wtrealm` and `client_id` are not compatible?

lib/auth0/api/authentication_endpoints.rb
```diff
@@ -215,12 +215,22 @@ def samlp_url(connection = UP_AUTH)
...
       # @see https://auth0.com/docs/api/authentication#accept-request35
       # @param connection [string] Connection to use; empty to show all
+      # @param options [hash] Extra options; supports wtrealm, wctx, wreply
       # @return [url] WS-Federation URL
-      def wsfed_url(connection = UP_AUTH)
+      def wsfed_url(connection = UP_AUTH, options = {})
         request_params = {
-          whr: connection
+          whr: connection,
+          wtrealm: options[:wtrealm],
```
lbalmaceda: Please add a test that asserts these properties are not being sent in the body when "null" or missing

lib/auth0/api/authentication_endpoints.rb
joshcanhelp: https://auth0.com/docs/api/authentication#accept-request35

lib/auth0/api/authentication_endpoints.rb
joshcanhelp: This just generates a URL, no HTTP requests

PR #126, opened 2018-08-01
Explicitly require JSON dependency

jgaskins:
This module depends on JSON but doesn't load it. This commit makes that dependency explicit so the application doesn't have to load it.

In all fairness, it's pretty likely that the app _does_ load the `json` gem/stdlib (which is why this has been able to go unnoticed), but we should always load our own dependencies. :-)


jgaskins:
Thanks @joshcanhelp!

PR #129, opened 2018-09-14
Add Client Credentials grant

### LINKED ISSUE #86: Obtain APIv2 Token
Greetings - it is possible to obtain an APIv2 token using the ruby library?  Basically, I want to call a method that makes the request documented here:

https://auth0.com/docs/api/management/v2/tokens#automate-the-process

Is there a way to do this currently with this library?

joshcanhelp:
This PR adds a method to [get an API access token with a Client Credentials grant](https://auth0.com/docs/api/management/v2/tokens#get-a-token-for-production) is no token is passed directly to `Auth0::Client` during instantiation. 

**Specific changes:**

- Add `Auth0::Api::AuthenticationEndpoints::api_token` method that accepts an `audience` (defaults to the Management API audience for the `domain` used), [performs a Client Credentials grant](https://auth0.com/docs/api/authentication#client-credentials), and returns a hash of what was returned. 
- Add a check during `Auth0::Client` instantiation to call this added method if no token was passed
- Add unit tests for new functionality and existing functionality

**Usage:**

To start using this new functionality, leave out `token` and `access_token` and add `client_secret` when instantiating `Auth0::Client`:

```ruby
# Set the Auth0 API client
def auth0_client
  # This assumes you're using env variables.
  @auth0_client ||= Auth0::Client.new(
    # REMOVE: token: ENV['AUTH0_RUBY_API_TOKEN'],
    # REMOVE: access_token: ENV['AUTH0_RUBY_API_TOKEN'],
    client_id:  ENV['AUTH0_RUBY_CLIENT_ID'],
    client_secret: ENV['AUTH0_RUBY_CLIENT_SECRET'],
    domain: ENV['AUTH0_RUBY_DOMAIN']
  )
end
```

`Auth0::Client.new` will see that you did not pass in a token and that there is a `client_secret` available and make a call to `oauth/token` to retrieve an API token for your Application. This will require that:

- the Application has "Client Credentials" selected in the Applications > Your app > Show Advanced Settings > Grant Types tab.
- the Application is authorized for the API in the APIs > Auth0 Management API > Machine to Machine Applications tab with scopes set to the Management API actions you want to perform.

You can also use the method directly for other APIs:

```ruby
require 'auth0'
auth0_client = Auth0::Client.new(
  client_id: ENV['AUTH0_RUBY_CLIENT_ID'],
  client_secret: ENV['AUTH0_RUBY_CLIENT_SECRET'],
  domain: ENV['AUTH0_RUBY_DOMAIN'],
)
token_result = auth0_client.api_token('your-audience-goes-here')
abort token_result.inspect
```
**Please note:** the API token should only be fetched once and cached *securely* for re-use before the expiration time. Each call to `api_token` or instantiation of `Auth0::Client.new` will make an HTTP call to get a new token. 

Closes #86


lib/auth0/api/authentication_endpoints.rb
```diff
@@ -7,7 +7,20 @@ module AuthenticationEndpoints
...
+      # Request an API access token using a Client Credentials grant
+      # @see https://auth0.com/docs/api-auth/tutorials/client-credentials
+      def api_token(audience = "https://#{@domain}/api/v2/")
+        request_params = {
+          grant_type: 'client_credentials',
+          client_id: @client_id,
+          client_secret: @client_secret,
+          audience: audience
+        }
+        post('/oauth/token', request_params)
```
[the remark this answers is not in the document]
joshcanhelp: > What kind of errors can be raised from this method?

Not totally sure. Is it important to list them? It uses these under the hood:

https://github.com/rest-client/rest-client which uses:
https://ruby-doc.org/stdlib-2.5.1/libdoc/net/http/rdoc/Net/HTTP.html

> Any reason the request params aren't inlined into the method call?

None of them except `audience` need to be set. Also, just following the example of the rest of the methods here. Would that look like:

```ruby
def api_token(
  client_id: @client_id,
  client_secret: @client_secret,
  audience: "https://#{@domain}/api/v2/"
)
  request_params = {
    grant_type: 'client_credentials',
    client_id: client_id,
    client_secret: client_secret,
    audience: audience
  }
  post('/oauth/token', request_params)
end
```

spec/lib/auth0/api/authentication_endpoints_spec.rb
```diff
@@ -44,12 +79,12 @@
...
     it "is expected to make post request to '/oauth/token'" do
       allow(@instance).to receive(:post).with(
-        '/oauth/token', client_id: @instance.client_id, client_secret: nil, grant_type: 'authorization_code',
+        '/oauth/token', client_id: @instance.client_id, client_secret: @instance.client_secret, grant_type: 'authorization_code',
                         connection: 'facebook', code: 'code', scope: 'openid', redirect_uri: 'uri'
       )
                                         .and_return('user_tokens' => 'UserToken')
       expect(@instance).to receive(:post).with(
-        '/oauth/token', client_id: @instance.client_id, client_secret: nil, grant_type: 'authorization_code',
+        '/oauth/token', client_id: @instance.client_id, client_secret: @instance.client_secret, grant_type: 'authorization_code',
```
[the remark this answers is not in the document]
joshcanhelp: > 1. Above you break the params across multiple lines, but here you leave it inlined

I didn't want to go too crazy with the non-functional changes. Tons of cluttered in-line stuff there already, this is just one line. 

> 2. These all seem to be doing the same exact things with different grant types or making slightly different assertions.

This one modification is for a soon-to-be-deprecated method that does something similar but not quite. This new method replaces it:

https://github.com/auth0/ruby-auth0/blob/master/lib/auth0/api/authentication_endpoints.rb#L15

lib/auth0/api/authentication_endpoints.rb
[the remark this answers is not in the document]
joshcanhelp: Pushing that change, would like to go this direction in general and need to start somewhere.

lib/auth0/api/authentication_endpoints.rb
```diff
@@ -7,7 +7,27 @@ module AuthenticationEndpoints
...
+        client_secret: @client_secret,
+        audience: "https://#{@domain}/api/v2/"
+      )
+        request_params = {
+          grant_type: 'client_credentials',
+          client_id: client_id,
+          client_secret: client_secret,
+          audience: audience
+        }
+        post('/oauth/token', request_params)
```
machuga: Each error type doesn't need accounted for by name, but it's up to you whether you'd like to catch errors here and return a single custom one or two.

lib/auth0/api/authentication_endpoints.rb
```diff
@@ -7,7 +7,27 @@ module AuthenticationEndpoints
...
 
+      # Request an API access token using a Client Credentials grant
+      # @see https://auth0.com/docs/api-auth/tutorials/client-credentials
+      # @param audience [string] API audience to use
+      # @return [json] Returns the API token
+      # def api_token(audience = "https://#{@domain}/api/v2/")
+      def api_token(
+        client_id: @client_id,
+        client_secret: @client_secret,
+        audience: "https://#{@domain}/api/v2/"
```
machuga: Wrapping this below in a `URI(audience).to_s` like you have in other gems is a good way to get some extra runtime guarantees, but is optional.

joshcanhelp: The `audience` param could be anything, doesn't have to be a URI:

screenshot 2018-09-20 09 48 28

lib/auth0/api/authentication_endpoints.rb
joshcanhelp: Though I think that's probably a good idea at some point, it's a breaking change to add that to all methods here and I don't want this to be any different than the others in that respect. I'll keep that in mind for the next major.

ambethia:
Really looking forward to this PR!

Issue #135, opened 2018-10-10
Issues building the documentation


gurbraj:
Hello,

When I run 
`$ bundle exec rake documentation `
I get
```
bundle exec rake documentation        
rake aborted!
Don't know how to build task 'documentation' (see --tasks)
```
When I list my tasks using `--tasks`, documentation does not show up.

Any ideas?

When the local documentation is built, will it be the same as what is on https://www.rubydoc.info/github/auth0/ruby-auth0/Auth0Client? It seems quite incomplete...

joshcanhelp:
Sorry for the trouble here and thank you for the report!

This is the up-to-date docs for the latest version:

https://www.rubydoc.info/gems/auth0/4.5.0/Auth0/Api/AuthenticationEndpoints

I'll investigate why the link you posted here is not up-to-date. It mentions `master` on there but I'm not seeing the latest auth endpoint docs added. 

I'm seeing the task on my end and it runs without any issues (a few warning that need cleaning up). It creates an up-to-date version in the `/docs` folder for this gem. [The task is initiated here](https://github.com/auth0/ruby-auth0/blob/master/Rakefile#L16). It might be a problem with how that task is structured so you can try one of the fixes [here](https://stackoverflow.com/questions/5689653/why-is-my-custom-rake-task-in-lib-tasks-not-discovered-in-rails-3). If one of those work, let me know or create a PR and I'll test it out. 

Also, can you try @jaimietanna's fix above (#136) and see if that makes a difference?

Thank you!

joshcanhelp:
Closed for no activity

PR #136, opened 2018-10-13
Add missing redcarpet dependency

jamietanna:
While investigating #135, I noticed that the build requires the
redcarpet gem, otherwise we get the error:

    [error]: Missing 'redcarpet' gem for Markdown formatting. Install it
    with `gem install redcarpet`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,7 @@ gemspec
 group :development do
   gem 'terminal-notifier-guard', require: false unless ENV['TRAVIS']
   gem 'coveralls', require: false
+  gem 'redcarpet', require: false
   gem 'rubocop', require: false
   gem 'yard', require: false
 end


joshcanhelp:
@jamietanna - Thanks so much for digging in on this!

Are you seeing that error when you run `bundle exec rake documentation`? I'm not seeing that error locally or in CI. I'm also not seeing this used anywhere in this gem.

jamietanna:
Interesting - I'm seeing that when running:

```
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]
Bundler version 1.16.6
```

joshcanhelp:
Is this within the gem or in your own project? I'm on the same version of Rails but downgraded Bundler to 1.16.2 to match the CI. It's curious because, like I said, I don't see anything in the gem that uses redcarpet or Markdown in general.

jamietanna:
Sorry, this is within just this project. I'll verify against different versions of Ruby in Docker, to see if there's anything I've broken on the host.

joshcanhelp:
No activity ... happy to re-open if you discover that this makes a difference.

Issue #92, opened 2017-08-15
Client specific post email verification redirect URL problem in V2


tadejm:
Greetings fellow Auth0 geeks,

I have an existing Auth0 account with multiple clients and [conditional post email verification redirects](https://auth0.com/docs/email/custom#custom-redirect) specified through the Auth0 dashboard.
When triggering a resend verification email job via the Ruby V2 client, the redirect to URL points to the default URL, not the one specified for the given `client_id`.

This is not the behavior in the V1 implementation where the `client_id` is correctly applied and the redirect follows to the specified URL.

Looking at the docs for [Send a "verify email address" email](https://auth0.com/docs/api/management/v2#!/Jobs/post_verification_email) a `client_id` body attribute is accepted, where as [the Ruby implementation](https://github.com/auth0/ruby-auth0/blob/master/lib/auth0/api/v2/jobs.rb#L42-L49) (to my understanding) does not apply the `client_id`:

```ruby
def send_verification_email(user_id)
  raise Auth0::InvalidParameter, 'Must specify a user id' if user_id.to_s.empty?
  request_params = {
    user_id: user_id
  }
  path = "#{jobs_path}/verification-email"
  post(path, request_params)
end
```

Naively, should we add an optional `client_id` argument and add it to the `request_params` if the value is present?

Thanks for your feedback

unkw:
I have the same problem , my {application.callback_domain} is empty string and get error as: {"name":"BadRequestError","code":"invalid_result_url","description":"invalid result url: /confirmation-email?...20once&success=false","statusCode":400}

joshcanhelp:
Apologies for the long delay in addressing this. Thank you for the very complete report here. As soon as #140 is merged in, I'll add that parameter and release it with [v4.6.0](https://github.com/auth0/ruby-auth0/milestone/2).

tadejm:
Awesome! Thanks @joshcanhelp :purple_heart:

PR #140, opened 2018-10-22
Added export users to jobs endpoint

raldred:
Add `.export_users` to `Auth0::Api::V2::Jobs`

### DIFF
--- a/lib/auth0/api/v2/jobs.rb
+++ b/lib/auth0/api/v2/jobs.rb
@@ -48,7 +48,7 @@ module Auth0
             connection_id: options.fetch(:connection_id, nil),
             format: options.fetch(:format, nil),
             limit: options.fetch(:limit, nil),
-            fields: fields_for_export(options.fetch(:connection_id, nil))
+            fields: fields_for_export(options.fetch(:fields, nil))
           }
           path = "#{jobs_path}/users-exports"
           post(path, request_params)
--- a/lib/auth0/api/v2/jobs.rb
+++ b/lib/auth0/api/v2/jobs.rb
@@ -75,6 +75,10 @@ module Auth0
           @jobs_path ||= '/api/v2/jobs'
         end
 
+        # Map array of field names for export to array of objects
+        # @param fields [array] Field names to be included in the export
+        
+        # @return [array] Returns the fields mapped as array of objects for the export_users endpoint
         def fields_for_export(fields)
           return nil if fields.nil? || fields.empty?
           fields.map { |field| { name: field } }
--- a/lib/auth0/api/v2/jobs.rb
+++ b/lib/auth0/api/v2/jobs.rb
@@ -34,6 +34,26 @@ module Auth0
           post_file(path, request_params)
         end
 
+        # Export all users to a file using a long running job.
+        # @see https://auth0.com/docs/api/v2#!/Jobs/post_users_exports
+        # @param options [hash] The options used to configure the output file.
+        #   * :connection_id [string] The connection id of the connection from which users will be exported.
+        #   * :format [string] The format of the file. Valid values are: "json" and "csv".
+        #   * :limit [integer] Limit the number of users to export.
+        #   * :fields [array] A list of fields to be included in the CSV. If omitted, a set of predefined fields will be exported.
+        #
+        # @return [json] Returns the job status and properties.
+        def export_users(options = {})
+          request_params = {
+            connection_id: options.fetch(:connection_id, nil),
+            format: options.fetch(:format, nil),
+            limit: options.fetch(:limit, nil),
+            fields: fields_for_export(options.fetch(:connection_id, nil))
+          }
+          path = "#{jobs_path}/users-exports"
+          post(path, request_params)
+        end
+
         # Send an email to the specified user that asks them to click a link to verify their email address.
         # @see https://auth0.com/docs/api/v2#!/Jobs/post_verification_email
         # @param user_id [string] The user_id of the user to whom the email will be sent.
@@ -53,6 +73,11 @@ module Auth0
         # Jobs API path
         def jobs_path
           @jobs_path ||= '/api/v2/jobs'
+        end
+
+        def fields_for_export(fields)
+          return nil if fields.nil? || fields.empty?
+          fields.map { |field| { name: field } }
         end
       end
     end
--- a/spec/integration/lib/auth0/api/v2/api_jobs_spec.rb
+++ b/spec/integration/lib/auth0/api/v2/api_jobs_spec.rb
@@ -44,6 +44,29 @@ describe Auth0::Api::V2::Jobs do
       end
     end
 
+    describe '.export_users and .get_job' do
+      let(:connection_id) do
+        client.connections
+              .find do |connection|
+          connection['name'].include?(Auth0::Api::AuthenticationEndpoints::UP_AUTH)
+        end['id']
+      end
+      let(:exported_users) { client.export_users }
+      it do
+        expect(exported_users).to include(
+          'connection' => Auth0::Api::AuthenticationEndpoints::UP_AUTH,
+          'status' => 'pending',
+          'type' => 'users_export'
+        )
+      end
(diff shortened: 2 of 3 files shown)


joshcanhelp:
Thank you @raldred! At first glance, this looks great. I'll pull it down and test it later this week 👍

spec/integration/lib/auth0/api/v2/api_jobs_spec.rb
```diff
@@ -44,6 +44,29 @@
       end
     end
 
+    describe '.export_users and .get_job' do
+      let(:connection_id) do
```
joshcanhelp: You're getting the `:connection_id` here but not using it anywhere.

spec/integration/lib/auth0/api/v2/api_jobs_spec.rb
```diff
@@ -44,6 +44,29 @@
...
+        client.connections
+              .find do |connection|
+          connection['name'].include?(Auth0::Api::AuthenticationEndpoints::UP_AUTH)
+        end['id']
+      end
+      let(:exported_users) { client.export_users }
+      it do
+        expect(exported_users).to include(
+          'connection' => Auth0::Api::AuthenticationEndpoints::UP_AUTH,
```
joshcanhelp: ... which is why this test will fail if you move it out of the `skip` block and run it.

spec/integration/lib/auth0/api/v2/api_jobs_spec.rb
```diff
@@ -44,6 +44,29 @@
...
+          'status' => 'pending',
+          'type' => 'users_export'
+        )
+      end
+      let(:export_job_id) { exported_users['id'] }
+      it do
+        expect(client.get_job(export_job_id)).to include(
+          'connection' => Auth0::Api::AuthenticationEndpoints::UP_AUTH, 'type' => 'users_export', 'id' => export_job_id
```
joshcanhelp: This one fails as well.

lib/auth0/api/v2/jobs.rb
```diff
@@ -54,6 +74,11 @@ def send_verification_email(user_id)
         def jobs_path
           @jobs_path ||= '/api/v2/jobs'
         end
+
+        def fields_for_export(fields)
```
joshcanhelp: Needs a docblock

spec/integration/lib/auth0/api/v2/api_jobs_spec.rb
```diff
@@ -44,6 +44,29 @@
...
+      let(:exported_users) { client.export_users }
+      it do
+        expect(exported_users).to include(
+          'connection' => Auth0::Api::AuthenticationEndpoints::UP_AUTH,
+          'status' => 'pending',
+          'type' => 'users_export'
+        )
+      end
+      let(:export_job_id) { exported_users['id'] }
+      it do
```
joshcanhelp: Need a test description here

spec/integration/lib/auth0/api/v2/api_jobs_spec.rb
```diff
@@ -44,6 +44,29 @@
...
 
+    describe '.export_users and .get_job' do
+      let(:connection_id) do
+        client.connections
+              .find do |connection|
+          connection['name'].include?(Auth0::Api::AuthenticationEndpoints::UP_AUTH)
+        end['id']
+      end
+      let(:exported_users) { client.export_users }
+      it do
```
joshcanhelp: Need a test description here

spec/integration/lib/auth0/api/v2/api_jobs_spec.rb
```diff
@@ -44,6 +44,29 @@
...
     end
 
+    describe '.export_users and .get_job' do
+      let(:connection_id) do
+        client.connections
+              .find do |connection|
+          connection['name'].include?(Auth0::Api::AuthenticationEndpoints::UP_AUTH)
+        end['id']
+      end
+      let(:exported_users) { client.export_users }
```
joshcanhelp: Use the `:connection_id` here

spec/lib/auth0/api/v2/jobs_spec.rb
```diff
@@ -24,7 +24,12 @@
     it { expect { @instance.import_users('', 'connnection_id') }.to raise_error('Must specify a valid file') }
     it { expect { @instance.import_users('users', '') }.to raise_error('Must specify a connection_id') }
   end
-
+  context '.export_users' do
+    it { expect(@instance).to respond_to(:export_users) }
+    it 'expect client to send post to /api/v2/jobs/users-exports' do
```
joshcanhelp: This is testing the default options but there should also be one that tests passed-in parameters, making sure they get mapped correctly ... particularly `fields` since that's processed in the endpoint method.

lib/auth0/api/v2/jobs.rb
```diff
@@ -54,6 +74,11 @@ def send_verification_email(user_id)
...
+          return nil if fields.nil? || fields.empty?
+          fields.map { |field| { name: field } }
```
joshcanhelp: There is also an `export_as` property that can be set.

joshcanhelp:
@raldred - Just checking in ... are you able to make these changes? Thank you!

joshcanhelp:
@raldred - Thanks again for your work here. We're going to release this week and would like to get these changes in if you're able to make the suggested updates.

raldred:
Hey Josh, sorry for the delay. I've been away, I will pick this up tomorrow.

joshcanhelp:
@raldred - Last check-in here. Are you able to revise this PR or should I close it out? I'm waiting on this PR to fix #92. 

Thank you either way!

spec/integration/lib/auth0/api/v2/api_jobs_spec.rb
raldred: Sorry I don't really follow. This test seems to use the 'spec_helper' which does not include the Credentials module so I actually get the following error.

```


NameError:
  undefined local variable or method `v2_creds' for #<RSpec::ExampleGroups::Auth0ApiV2Jobs after(:context) hook>
# ./spec/integration/lib/auth0/api/v2/api_jobs_spec.rb:89:in `block (2 levels) in <top (required)>'
 1/6 |=== 16 ====>                                                            |  ETA: 00:00:00
Finished in 0.03837 seconds (files took 0.51953 seconds to load)
1 example, 1 failure, 1 error occurred outside of examples

Failed examples:

rspec ./spec/integration/lib/auth0/api/v2/api_jobs_spec.rb:32 
```

raldred: If I use MODE=full I get the following error for all tests in this file, not just the one I added.
```
1) Auth0::Api::V2::Jobs.import_users and .get_job
     Failure/Error: raise InvalidApiNamespace, 'API namespace must supply an API domain' if @domain.to_s.empty?

     Auth0::InvalidApiNamespace:
       API namespace must supply an API domain
```

joshcanhelp: @raldred - Thanks for sticking with this!

We've made a few changes to the testing suite that should make this work a bit more smoothly. On your local copy, rebase with our `master` and run:

```bash
$ bundle exec rake test
```

I just ran that process on your branch locally and everything passes in it's current state (assuming all of your changes are pushed). I moved your tests out of the skip block and they complete just fine in the state they're in. 

One thing to note ... you'll need to uncomment [this line](https://github.com/auth0/ruby-auth0/blob/master/spec/spec_helper.rb#L25) to allow HTTP connections temporarily when you move the tests out of the skip block. Otherwise VCR (records the HTTP calls so the tests ran faster and don't hammer our servers) will stop the call. 

Try the above and, if it works, just add the test descriptions and we're good to go. If not, ditch the integration tests and I'll handle them on my end. This file needs work generally and we need a CONTRIBUTING.md file to explain all of this stuff. 

Thanks again!

PR #143, opened 2018-10-26
Improve the test suite

joshcanhelp:
### Changes

- Remove `full` ENV var
- Add `test` Rake task
- Combine all spec helper files
- Install Codecov


spec/spec_helper.rb
```diff
@@ -14,25 +29,30 @@
   config.filter_sensitive_data('API_TOKEN') { ENV['MASTER_JWT'] }
 end
 
-mode = ENV['MODE'] || 'unit'
```
joshcanhelp: Removed the test mode now that we're using VCR

spec/spec_helper.rb
```diff
@@ -14,25 +29,30 @@
...
-
 $LOAD_PATH.unshift File.expand_path('..', __FILE__)
 $LOAD_PATH.unshift File.expand_path('../../lib', __FILE__)
 
-require 'rspec'
```
joshcanhelp: Moved up

spec/spec_helper.rb
```diff
@@ -14,25 +29,30 @@
...
+  raise e if elapsed_time >= time
+  sleep increment
+  wait(time, increment, elapsed_time + increment, &block)
+end
+
 def entity_suffix
   'rubytest'
 end
 
-puts "Entity suffix is #{entity_suffix}"
```
joshcanhelp: Moving test mode-specific code

Gemfile
```diff
@@ -13,4 +13,5 @@ end
 group :test do
   gem 'webmock', require: false
   gem 'vcr', require: false
+  gem 'codecov', :require => false
```
lbalmaceda: why the arrow?

spec/spec_helper.rb
```diff
@@ -6,6 +20,7 @@
 
 require 'vcr'
 VCR.configure do |config|
+  # Uncomment the line below to record new cassettes
```
lbalmaceda: what does that mean? should it be clarified?

spec/spec_helper.rb
lbalmaceda: VCR?

spec/spec_helper.rb
lbalmaceda: down you mean? :trollface:

spec/spec_helper.rb
```diff
@@ -14,25 +29,30 @@
...
-require 'rspec'
-require 'rack/test'
-require 'faker'
-require 'auth0'
-
 Dir['./lib/**/*.rb'].each { |f| require f }
 Dir['./spec/support/**/*.rb'].each { |f| require f }
 Dir['./spec/support/*.rb'].each { |f| require f }
 
+require 'rspec'
```
lbalmaceda: cosmetic but, shouldn't this be on the top of the file instead of flying here?

joshcanhelp: Actually no ... `config.include Credentials` needs to happen after the file requires

PR #145, opened 2018-11-09
Add ttl_sec argument to post_email_verification request.

digitaldawn:
### Changes

Add ttl_sec argument to post_email_verification request. 

### References

The post_email_verification endpoint supports a ttl_sec parameter which this gem does not currently support. See: https://auth0.com/docs/api/management/v2#!/Tickets/post_email_verification

### Testing

The tickets spec was adjusted to include the relevant change here. Admittedly this is blind as I couldn't for the life of me figure our how to run the specs. If anyone can point me to some documentation or guide me through this then that would be great - otherwise I assume this is a relatively non-impactful change and shouldn't break anything. 

### Checklist

### DIFF
--- a/lib/auth0/api/v2/tickets.rb
+++ b/lib/auth0/api/v2/tickets.rb
@@ -22,7 +22,7 @@ module Auth0
           request_params = {
             user_id: user_id,
             result_url: result_url,
-            ttl_sec: ttl_sec
+            ttl_sec: ttl_sec.is_a?(Integer) ? ttl_sec : nil
           }
           post(path, request_params)
         end
--- a/lib/auth0/api/v2/tickets.rb
+++ b/lib/auth0/api/v2/tickets.rb
@@ -9,16 +9,20 @@ module Auth0
         # @see https://auth0.com/docs/api/v2#!/Tickets/post_email_verification
         # @param user_id [string] The user_id of for which the ticket is to be created.
         # @param result_url [string] The user will be redirected to this endpoint once the ticket is used.
+        # @param ttl_sec [integer] The ticket's lifetime in seconds starting from the moment of creation.
+        # After expiration, the ticket cannot be used to verify the user's email. If not specified or if
+        # you send 0, the Auth0 default lifetime of five days will be applied
         #
         # @return [json] Returns the created ticket url.
-        def post_email_verification(user_id, result_url: nil)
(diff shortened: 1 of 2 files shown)


joshcanhelp:
Thank you @digitaldawn, I'll take a look as soon as I can!

lib/auth0/api/v2/tickets.rb
```diff
@@ -9,16 +9,20 @@ module Tickets
...
+        def post_email_verification(user_id, result_url: nil, ttl_sec: nil)
           if user_id.to_s.empty?
             raise Auth0::InvalidParameter, 'Must supply a valid user id to post an email verification'
           end
           path = "#{tickets_path}/email-verification"
           request_params = {
             user_id: user_id,
-            result_url: result_url
+            result_url: result_url,
+            ttl_sec: ttl_sec
```
joshcanhelp: Since the API is looking for an integer, I think it would be worth it to check with something like:

```ruby
ttl_sec: ttl_sec if ttl_sec.is_a? Integer
```

... or similar. The API will return a `400` so we can prevent that before an HTTP call is made.

spec/lib/auth0/api/v2/tickets_spec.rb
```diff
@@ -7,7 +7,8 @@
   context '.post_email_verification' do
     it { expect(@instance).to respond_to(:post_email_verification) }
     it 'expect client to send post to /api/v2/tickets/email-verification with body' do
-      expect(@instance).to receive(:post).with('/api/v2/tickets/email-verification', user_id: 'user_id', result_url: nil)
+      expect(@instance).to receive(:post).with('/api/v2/tickets/email-verification', user_id: 'user_id',
```
joshcanhelp: Probably want one more for a non-nil `ttl_sec` ... and another to confirm what happens if a non-Integer is used for `ttl_sec`

digitaldawn:
Hi @joshcanhelp, apologies I should have been clearer with the test scenario. I was able to run the tests but they were all failing as it was looking for credentials (and I wasn't sure to get these).
I managed to pull them out of the VCR cassettes so can now run the tests. Will get this touched up.

digitaldawn:
I think I've now added the changes you requested :)

joshcanhelp:
Thanks @digitaldawn! Looks good at first glance here, I'll review as soon as I can and get this merged.

Issue #147, opened 2018-11-29
Unable to initialize Auth0Api


reckerswartz:
### Description
Unable to initialize Auth0Api
for  [obtain_access_token](https://www.rubydoc.info/gems/auth0/4.5.0/Auth0%2FApi%2FAuthenticationEndpoints:obtain_access_token) 

tried various options include eg. Auth0Api.initialize(options = {})
Auth0Api.new(options = {})

### Environment

* Ruby Auth0 version: 4.5.0
* Ruby version: 2.4.4
* Rails version (if applicable): 5.1

joshcanhelp:
Sorry for the trouble here @reckerswartz. We're working on improving the composition of these modules now but right now, to access that method, you can do one of two things.

You can include the necessary modules in your class and define the necessary `client_headers` method and class properties in your class like this (using a controller as an example):

```ruby
class YourController < ApplicationController

  before_action :set_auth0_vars, only: %i[process_access_token]

  include Auth0::Api::AuthenticationEndpoints
  include Auth0::Mixins::HTTPProxy

  def token_page
    render 'token_page'
  end

  def process_access_token
    token_params = params[:token_params]
    response = obtain_access_token(token_params[:access_token], token_params[:connection])
    # ...
  end

  # ...

  private

  def set_auth0_vars
    @domain = ENV['AUTH0_RUBY_DOMAIN']
    @client_id = ENV['AUTH0_RUBY_CLIENT_ID']
    @client_secret = ENV['AUTH0_RUBY_CLIENT_SECRET']
    @base_uri = "https://#{@domain}"
    @headers = client_headers
  end

  def client_headers
    client_info = JSON.dump(name: 'ruby-auth0', version: Auth0::VERSION)
    {
      'Content-Type' => 'application/json',
      'User-Agent' => "Ruby/#{RUBY_VERSION}",
      'Auth0-Client' => Base64.urlsafe_encode64(client_info)
    }
  end
end
```

Or you can use the complete SDK client but you'll have to pass in a dummy Management token to get it to instantiate. Similar example as the above:

```ruby
require 'auth0'
class YourController < ApplicationController

  def token_page
    render 'token_page'
  end

  def process_access_token
    token_params = params[:token_params]
    response = auth0_client.obtain_access_token(token_params[:access_token], token_params[:connection])
    # ...
  end

  # ...

  private

  def auth0_client
    @client_id = ENV['AUTH0_RUBY_CLIENT_ID']
[...]

PR #149, opened 2018-12-03
Add telemetry

joshcanhelp:
### Changes

Add module `Auth0::Mixins::Headers` to handle telemetry headers more easily; add environmental telemetry (Ruby and Rails versions). This will make it easier to use `Auth0::Api::AuthenticationEndpoints` on its own like so:

```ruby
class StaticPagesController < ApplicationController

  include Auth0::Api::AuthenticationEndpoints
  include Auth0::Mixins::HTTPProxy
  include Auth0::Mixins::Headers

  before_action :set_auth0_vars, only: %i[process_login_ro]

  def process_login_ro
    login_ro = params[:login_ro]
    response = login_with_resource_owner(login_ro[:email], login_ro[:password])
  end

  private

  def set_auth0_vars
    @domain = ENV['AUTH0_RUBY_DOMAIN']
    @client_id = ENV['AUTH0_RUBY_CLIENT_ID']
    @client_secret = ENV['AUTH0_RUBY_CLIENT_SECRET']
    @base_uri = "https://#{@domain}"
    @headers = client_headers
  end
end
```

### References

This addresses an internal telemetry requirement while also making it easier to [handle situations like this](https://github.com/auth0/ruby-auth0/issues/147). 

### Testing

### Checklist


spec/integration/lib/auth0/auth0_client_spec.rb
```diff
@@ -62,13 +62,12 @@
   context 'client headers' do
     let(:client) { Auth0::Client.new(v2_credentials.merge(access_token: 'abc123', domain: 'myhost.auth0.com')) }
     let(:headers) { client.headers }
-
-    let(:base64_token) do
-      Base64.urlsafe_encode64('{"name":"ruby-auth0","version":"' + Auth0::VERSION + '"}')
+    let(:telemetry) do
```
machuga: Can you update this to use `{}` syntax, as this repo seems to use more of a Weirich-style block syntax? (meaning it will also match the above)

joshcanhelp: @machuga - Fixed up!

PR #150, opened 2018-12-05
Add refresh token method and unit tests

### LINKED ISSUE #111: Cannot refresh access token with refresh token.
It would appear the `login` function in `authentication_endpoints.rb`, which POSTs to `oauth/token` does not appear to allow for an optional refresh token field. 

According to the Auth0 docs, this endpoint is also used to get a new access token via refresh token if one is available (using `refresh_token` in the `grant_field` option as well as providing the `refresh_token` as a top level param). 

To me, this means that there is no way to use a refresh token to get a new access token with this API wrapper.

Is this correct?

joshcanhelp:
### Changes

- Add `Auth0::Api::AuthenticationEndpoints.exchange_refresh_token` method.
- Remove requirement for `:redirect_uri` param in the `exchange_auth_code_for_tokens`method

### References

Closes #111 

### Testing

Also add a test for `exchange_auth_code_for_tokens` with custom params. 

* [ ] This change adds integration test coverage (currently no integration tests for tokens)
* [x] This change has been tested on the latest version of Ruby (ruby-2.5.1)

### Checklist

### DIFF
--- a/lib/auth0/api/authentication_endpoints.rb
+++ b/lib/auth0/api/authentication_endpoints.rb
@@ -29,22 +29,48 @@ module Auth0
       # Get access and ID tokens using an Authorization Code.
       # @see https://auth0.com/docs/api/authentication#authorization-code
       # @param code [string] The authentication code obtained from /authorize
-      # @param redirect_uri [string] Url to redirect after authorization
-      # @return [json] Returns the access_token and id_token
+      # @param redirect_uri [string] URL to redirect to after authorization.
+      #   Required only if it was set at the GET /authorize endpoint
+      # @param client_id [string] Client ID for the Application
+      # @param client_secret [string] Client Secret for the Application.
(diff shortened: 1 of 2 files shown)


lib/auth0/api/authentication_endpoints.rb
```diff
@@ -49,6 +51,30 @@ def exchange_auth_code_for_tokens(
...
+        refresh_token,
+        client_id: @client_id,
+        client_secret: @client_secret
+      )
+        raise Auth0::InvalidParameter, 'Must provide a refresh token' if refresh_token.to_s.empty?
+        request_params = {
+          grant_type:    'refresh_token',
+          client_id:     client_id,
+          client_secret: client_secret,
+          refresh_token:  refresh_token
```
machuga: Alignment is off here. May be easiest to pull them all to 1 space from the colon?

Issue #154, opened 2019-01-15
No method to get all resource servers


philomory:
### Description

There's no method in the gem to get a list of all resource servers (e.g. to call `GET /api/v2/resource-servers`); corresponding methods exist for other endpoints such as Clients.

I have a PR ready to implement this change, but the contributing guidelines instruct to open an issue first, so, that's what I'm doing.

### Prerequisites

* [x] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
* [x] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md)
* [x] Did you check the documentation ([repo README](https://github.com/auth0/ruby-auth0/blob/master/README.md) or [Quickstart](https://auth0.com/docs/quickstart/backend/rails))?
* [x] Did you check the [Auth0 Community](https://community.auth0.com/)?
* [x] Are you reporting this to the correct repository? See also the [OmniAuth-Auth0 strategy](https://github.com/auth0/omniauth-auth0) for logging in with Rails.
* [x] Are there any related or duplicate [Issues](https://github.com/auth0/ruby-auth0/issues) or [PRs](https://github.com/auth0/ruby-auth0/pulls) for this issue?

### Environment

Please provide the following:

* Ruby Auth0 version: 4.6.0
* Ruby version: 2.4.4p296
* Rails version (if applicable): N/A
* Browser version (if applicable): N/A
* Additional gems that might be affecting your instance: N/A

### Reproduction

```
irb(main):001:0> connection = Auth0::Client.new(...)
irb(main):002:0> apis = connection.resource_servers
NoMethodError: undefined method `resource_servers' for #<Auth0::Client:0x0000000002e740a8>
Did you mean? resource_server
```

joshcanhelp:
Thank you @philomory! We'd be glad to accept a PR for this. A few things to keep in mind when you put this together:

- Unit tests should be included, go ahead and use one of the other endpoints as a template for that. 
- Integration tests are appreciated but not required. If you do tackle integration tests, you'll need to uncomment [this line](https://github.com/auth0/ruby-auth0/blob/master/spec/spec_helper.rb#L25) so VCR can make an HTTP recording. Once everything is working and passing, comment it out again and run the tests once more to make sure no HTTP connections are made.

philomory:
[...]

PR #155, opened 2019-01-19
Get all resource servers

philomory:
### Changes

Adds `Auth0::Api::V2::ResourceServers#resource_servers` method (and alias `get_resource_servers`) to fetch all Resource Servers in the tenant.

### References

Discussion in #154 

### Testing

* [ ] This change has been tested on the latest version of Ruby
(my testing was done on 2.4.4, not 2.6.0)

### Checklist

* [x] All existing and new tests complete without errors - specs pass, I can't run integration tests locally
* [ ] Rubocop passes on all added/modified files - they don't pass on `master` when I run them, either

### DIFF
--- a/spec/integration/lib/auth0/api/v2/api_resource_servers_spec.rb
+++ b/spec/integration/lib/auth0/api/v2/api_resource_servers_spec.rb
@@ -41,7 +41,6 @@ describe Auth0::Api::V2::ResourceServers do
         per_page: 1
       )
       expect(results.first).to equal(results.last)
-      expect(results.first).to eq(test_server)
     end
   end
 
--- a/spec/integration/lib/auth0/api/v2/api_resource_servers_spec.rb
+++ b/spec/integration/lib/auth0/api/v2/api_resource_servers_spec.rb
@@ -22,6 +22,29 @@ describe Auth0::Api::V2::ResourceServers do
     end
   end
 
+  describe '.resource_servers', vcr: true do
+    let(:resource_servers) do
+      client.resource_servers
+    end
+
+    it 'should return at least 1 result' do
+      expect(resource_servers.size).to be > 0
+    end
+
+    it 'should get the test server' do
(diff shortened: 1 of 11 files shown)


spec/integration/lib/auth0/api/v2/api_resource_servers_spec.rb
```diff
@@ -22,6 +22,29 @@
...
+      expect(resource_servers).to include(test_server)
+    end
+
+    it 'should return the first page of one result' do
+      results = client.resource_servers(
+        page: 0,
+        per_page: 1
+      )
+      expect(results.first).to equal(results.last)
+      expect(results.first).to eq(test_server)
```
joshcanhelp: This is not necessary, and is failing on my end when I run the integration tests on your branch. We're not pulling by any specific order here so the newest may (or may not) be the first one pulled.

spec/spec_helper.rb
```diff
@@ -22,7 +22,7 @@
 VCR.configure do |config|
   # Uncomment the line below to record new VCR cassettes.
   # When this is commented out, VCR will reject all outbound HTTP calls.
-  # config.allow_http_connections_when_no_cassette = true
+  config.allow_http_connections_when_no_cassette = true
```
joshcanhelp: Make sure to comment this back out so HTTP calls are not made.

spec/integration/lib/auth0/api/v2/api_resource_servers_spec.rb
philomory: I've removed the `expect(results.first).to eq(test_server)` line, which I _believe_ is the only one you objected to. If you want me to remove that whole test about paging, though, I can do that instead.

spec/spec_helper.rb
philomory: That's been commented-out again.

Issue #156, opened 2019-01-28
No method to update/patch resource servers


philomory:
### Description
There's no method in the gem to update a resource server (e.g. to call PATCH /api/v2/resource-servers/[id]); corresponding methods exist for other endpoints such as Clients.

I have a WIP PR for this.

### Prerequisites

* [x] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
* [x] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md)
* [x] Did you check the documentation ([repo README](https://github.com/auth0/ruby-auth0/blob/master/README.md) or [Quickstart](https://auth0.com/docs/quickstart/backend/rails))?
* [x] Did you check the [Auth0 Community](https://community.auth0.com/)?
* [x] Are you reporting this to the correct repository? See also the [OmniAuth-Auth0 strategy](https://github.com/auth0/omniauth-auth0) for logging in with Rails.
* [x] Are there any related or duplicate [Issues](https://github.com/auth0/ruby-auth0/issues) or [PRs](https://github.com/auth0/ruby-auth0/pulls) for this issue?

### Environment

Please provide the following:

* Ruby Auth0 version: 4.6.0
* Ruby version: 2.4.4p296
* Rails version (if applicable): N/A
* Browser version (if applicable): N/A
* Additional gems that might be affecting your instance: N/A

### Reproduction

```
irb(main):001:0> connection = Auth0::Client.new(...)
irb(main):002:0> connection.patch_resource_server(...)
NoMethodError: undefined method `patch_resource_servers' for #<Auth0::Client:0x00000000073974a0>
```

joshcanhelp:
@philomory - Appreciate the help with these! You are approved to proceed 😃

joshcanhelp:
Merged!

PR #157, opened 2019-01-31
Add patch resource server

philomory:
### Changes

Adds a `patch_resource_server` method that takes a resource server id and a hash of options. Includes spec tests that pass and tentative integration tests awaiting VCR cassettes.

### References

Discussion in #156.

### Testing

* [ ] This change has been tested on the latest version of Ruby - Tested on 2.4.4

### Checklist

### DIFF
--- a/lib/auth0/api/v2/resource_servers.rb
+++ b/lib/auth0/api/v2/resource_servers.rb
@@ -62,6 +62,17 @@ module Auth0
           delete(path)
         end
 
+        # Updates a resource server.
+        # @see https://auth0.com/docs/api/management/v2#!/Resource_Servers/patch_resource_servers_by_id
+        # @param id [string] The id or audience of the resource server to update.
+        # @param options [hash] The Hash options used to define the resource servers's properties.
+        def patch_resource_server(id, options)
+          raise Auth0::InvalidParameter, 'Must specify a resource server id' if id.to_s.empty?
+          raise Auth0::InvalidParameter, 'Must specify a valid body' if options.to_s.empty?
+          path = "#{resource_servers_path}/#{id}"
+          patch(path, options)
+        end
+        
         private
 
         # Resource Servers API path
--- a/lib/auth0/api/v2/resource_servers.rb
+++ b/lib/auth0/api/v2/resource_servers.rb
@@ -65,12 +65,12 @@ module Auth0
         # Updates a resource server.
         # @see https://auth0.com/docs/api/management/v2#!/Resource_Servers/patch_resource_servers_by_id
         # @param id [string] The id or audience of the resource server to update.
-        # @param options [hash] The Hash options used to define the resource servers's properties.
-        def patch_resource_server(id, options)
-          raise Auth0::InvalidParameter, 'Must specify a resource server id' if id.to_s.empty?
-          raise Auth0::InvalidParameter, 'Must specify a valid body' if options.to_s.empty?
+        # @param body [hash] The Hash options used to define the resource servers's properties.
+        def patch_resource_server(id, body)
+          raise Auth0::InvalidParameter, 'Must specify a resource server id or audience' if id.to_s.empty?
+          raise Auth0::InvalidParameter, 'Must specify a valid body' if body.to_s.empty?
           path = "#{resource_servers_path}/#{id}"
-          patch(path, options)
+          patch(path, body)
         end
         
         private
--- a/spec/integration/lib/auth0/api/v2/api_resource_servers_spec.rb
+++ b/spec/integration/lib/auth0/api/v2/api_resource_servers_spec.rb
@@ -98,22 +98,10 @@ describe Auth0::Api::V2::ResourceServers do
     end
   end
 
-    describe '.patch_client', vcr: true do
-    it 'should raise an error with a missing client_id' do
-      expect do
-        client.patch_resource_server('', token_lifetime: 654321)
-      end.to raise_error Auth0::InvalidParameter
-    end
-
-    it 'should raise an error with an empty body' do
-      expect do
-        client.patch_resource_server('__test_resource_server_id__', {})
-      end.to raise_error Auth0::InvalidParameter
-    end
-
+  describe '.patch_resource_server', vcr: true do
     it 'should update the resource server with the correct attributes' do
       expect(
-        client.patch_client(
+        client.patch_resource_server(
           test_server['id'],
           token_lifetime: 654321,
         )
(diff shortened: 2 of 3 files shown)


joshcanhelp:
Thanks @philomory! I'll review this ASAP.

lib/auth0/api/v2/resource_servers.rb
```diff
@@ -62,6 +62,17 @@ def delete_resource_server(resource_server_id)
           delete(path)
         end
 
+        # Updates a resource server.
+        # @see https://auth0.com/docs/api/management/v2#!/Resource_Servers/patch_resource_servers_by_id
+        # @param id [string] The id or audience of the resource server to update.
+        # @param options [hash] The Hash options used to define the resource servers's properties.
```
joshcanhelp: Let's change this parameter name to `body`. We typically use `options` for URL parameters.

lib/auth0/api/v2/resource_servers.rb
```diff
@@ -62,6 +62,17 @@ def delete_resource_server(resource_server_id)
...
+        def patch_resource_server(id, options)
+          raise Auth0::InvalidParameter, 'Must specify a resource server id' if id.to_s.empty?
```
joshcanhelp: Let's say:

> Must specify a resource server ID or audience

... since it can accept both.

philomory: Actually, on this topic, should we URI-escape this parameter, or except the caller to do it? Obviously irrelevant if the caller is passing an id, but if passing an audience, and the audience is of the common form `https://something.com/` then that needs to be URI-escaped before being added to the PATH string.

In my PR I left it up the to caller to URI escape audiences, since the `delete_resource_server`  method (the only existing method that puts an id-or-audience in the URI's path segment) doesn't handle URI escaping automatically, but an argument could be made to add it. Do you want me to go ahead and add that here, and to the delete method as well? Or, since that's theoretically a backwards-incompatible change (anyone previously using the delete method with an audience would be escaping before passing it in, and if we start escaping as well then it ends up double-escaped), should it wait for  a separate PR that could be part of 5.0.0?

lib/auth0/api/v2/resource_servers.rb
philomory: Done.

lib/auth0/api/v2/resource_servers.rb
joshcanhelp: That's a great question. 

I would prefer it be done in the SDK but none of the other methods I looked at for any of the endpoints do so I think we should be consistent here. We're putting together a roadmap for a major release on this library and I'll bring that up for discussion.

spec/lib/auth0/api/v2/resource_servers_spec.rb
```diff
@@ -73,4 +73,14 @@
...
     end
   end
+
+  context '.patch_resource_server' do
+    it { expect(@instance).to respond_to(:patch_resource_server) }
+    it 'is expected to send patch to /api/v2/resource_servers/1' do
+      expect(@instance).to receive(:patch).with('/api/v2/resource-servers/1', fields: 'fields')
+      expect { @instance.patch_resource_server('1', fields: 'fields') }.not_to raise_error
+    end
+    it { expect { @instance.patch_resource_server('', nil) }.to raise_error 'Must specify a resource server id' }
```
joshcanhelp: Tests are failing on this:

https://travis-ci.org/auth0/ruby-auth0/jobs/488774537#L565

I would just check for `Auth0::InvalidParameter` like the integration tests.

spec/integration/lib/auth0/api/v2/api_resource_servers_spec.rb
```diff
@@ -98,6 +98,33 @@
     end
   end
 
+    describe '.patch_client', vcr: true do
```
joshcanhelp: Just noticed ... indented too far

spec/integration/lib/auth0/api/v2/api_resource_servers_spec.rb
```diff
@@ -98,6 +98,33 @@
...
+    it 'should raise an error with a missing client_id' do
```
joshcanhelp: We can take out the integration specs that check for a raised error. The unit specs will cover just fine.

spec/integration/lib/auth0/api/v2/api_resource_servers_spec.rb
```diff
@@ -98,6 +98,33 @@
...
+      expect do
+        client.patch_resource_server('', token_lifetime: 654321)
+      end.to raise_error Auth0::InvalidParameter
+    end
+
+    it 'should raise an error with an empty body' do
```
joshcanhelp: Same

lib/auth0/api/v2/resource_servers.rb
philomory: It's worth noting that most endpoints don't accept anything in the the path segment of the URL that could ever possibly need URI escaping; the resource servers endpoint is the only one I can think of off the top of my head.

Just to play devils advocate, you could argue that adding this functionality isn't a compatibility-breaking change because the library did not previously document any support for accepting an audience rather than an id in these methods, so anyone who was passing one in was relying on undocumented behavior (of the library; the actual REST API does document this of course). That said, I'm not sure I buy that argument myself and am content to wait for 5.0.0 to introduce the automatic escaping.

spec/lib/auth0/api/v2/resource_servers_spec.rb
philomory: Made the requested change.

This actually reminds me of something I've been wondering about: why are `Auth0::MissingClientId` and `Auth0::MissingUserId` the only dedicated exceptions for a specific property? There's no `Auth0::MissingResourceServerId` or `Auth0::MissingConnectionId`, all other resources just use `Auth0::InvalidParameter` for a missing id, same as they do for other missing parameters or for present-but-invalid parameters.

spec/integration/lib/auth0/api/v2/api_resource_servers_spec.rb
philomory: Fixed this. Also changed the description string to say `.patch_resource_server` as it should.

lib/auth0/api/v2/resource_servers.rb
joshcanhelp: It's definitely the minority of cases, true. But also a consideration for URL parameters.

Every time I talk myself into changing something that shouldn't be a breaking change but is, there is inevitably someone affected! I think it makes more sense to address it gem-wide in a major. I appreciate the opportunity to talk it through, though.

PR #162, opened 2019-02-27
Add Client ID to send_verification_email

### LINKED ISSUE #92: Client specific post email verification redirect URL problem in V2

joshcanhelp:
### Changes

- Add Add Client ID to `Auth0::Api::V2::Jobs.send_verification_email`
- Remove skipped Jobs tests and add VCR cassettes

### References

Fixes #92 

### Testing

[Failing tests for new feature](https://travis-ci.org/auth0/ruby-auth0/builds/499524045)

* [x] This change has been tested on the latest version of Ruby (2.5.1)

### Checklist

### DIFF
--- a/lib/auth0/api/v2/jobs.rb
+++ b/lib/auth0/api/v2/jobs.rb
@@ -3,29 +3,29 @@ module Auth0
     module V2
       # Methods to use the jobs endpoints
       module Jobs
-        attr_reader :jobs_path
-
         # Retrieves a job. Useful to check its status.
         # @see https://auth0.com/docs/api/v2#!/Jobs/get_jobs_by_job_id
-        # @param job_id [string] The id of the job.
+        # @param job_id [string] The ID of the job.
         #
         # @return [json] Returns the job status and properties.
         def get_job(job_id)
           raise Auth0::InvalidParameter, 'Must specify a job id' if job_id.to_s.empty?
+
           path = "#{jobs_path}/#{job_id}"
           get(path)
         end
 
         # Imports users to a connection from a file using a long running job.
-        # Important: The documentation for the file format is at https://docs.auth0.com/bulk-import.
+        # Documentation for the file format: https://docs.auth0.com/bulk-import
         # @see https://auth0.com/docs/api/v2#!/Jobs/post_users_imports
         # @param users_file [file] A file containing the users to import.
-        # @param connection_id [string] The connection id of the connection to which users will be inserted.
+        # @param connection_id [string] Database connection ID to import to.
         #
         # @return [json] Returns the job status and properties.
         def import_users(users_file, connection_id)
           raise Auth0::InvalidParameter, 'Must specify a valid file' if users_file.to_s.empty?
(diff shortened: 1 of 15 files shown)


lib/auth0/api/v2/jobs.rb
```diff
@@ -76,11 +79,12 @@ def jobs_path
...
         # Map array of field names for export to array of objects
-        # @param fields [array] Field names to be included in the export
-        
-        # @return [array] Returns the fields mapped as array of objects for the export_users endpoint
+        # @param fields [array] Field names to be included in the export
+
+        # @return [array] Returns the fields mapped as array of objects for the export_users endpoint
         def fields_for_export(fields)
-          return nil if fields.nil? || fields.empty?
+          return nil if fields.to_s.empty?
```
luisrudge: why the change here?

lib/auth0/mixins/httpproxy.rb
```diff
@@ -21,6 +21,7 @@ module HTTPProxy
                    elsif method == :delete
                      call(:delete, url(safe_path), timeout, add_headers({params: body}))
                    elsif method == :post_file
+                     body.merge!(multipart: true)
```
luisrudge: why?

lib/auth0/api/v2/jobs.rb
joshcanhelp: Simplification ... I think it was suggested by the formatting tool `rubocop`

lib/auth0/mixins/httpproxy.rb
joshcanhelp: The import method was not working as it was written, which I found out when the tests were finally running (instead of being skipped).

joshcanhelp: The import method was not working as it was written, which I found out when the tests were finally running (instead of being skipped).

Issue #158, opened 2019-02-01
Not enough information in exception when Rate Limiting is encountered


philomory:
### Description

When rate limiting is encountered, an `Auth0::Unsupported` exception is raised, whose message only states `{"statusCode":429,"error":"Too Many Requests"}`. This is because the exception message is generated from only the HTTP response's status and body.

When the Auth0 API returns a 429 response, it includes details about the limits in HTTP response headers; specifically, as noted in [Rate Limits](https://auth0.com/docs/policies/rate-limits#http-response-headers), the `X-RateLimit-Limit`, `X-RateLimit-Remaining` and `X-RateLimit-Reset` headers.

It would be good to return this information to the caller when raising an exception due to rate limiting.

### Prerequisites

* [x] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
* [x] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md)
* [x] Did you check the documentation ([repo README](https://github.com/auth0/ruby-auth0/blob/master/README.md) or [Quickstart](https://auth0.com/docs/quickstart/backend/rails))?
* [x] Did you check the [Auth0 Community](https://community.auth0.com/)?
* [x] Are you reporting this to the correct repository? See also the [OmniAuth-Auth0 strategy](https://github.com/auth0/omniauth-auth0) for logging in with Rails.
* [x] Are there any related or duplicate [Issues](https://github.com/auth0/ruby-auth0/issues) or [PRs](https://github.com/auth0/ruby-auth0/pulls) for this issue?

### Environment

Please provide the following:

* Ruby Auth0 version: 4.6.0
* Ruby version: 2.4.4
* Rails version (if applicable): N/A
* Browser version (if applicable): N/A
* Additional gems that might be affecting your instance: N/A

### Reproduction

Easiest way to reproduce is to use a free/trial account with a low rate limit, and then repeatedly call `Auth0::Api::V2#logs` until an exception is raised.

philomory:
Actually, it might be generally a good idea to make all of the exceptions that `Auth0::Mixins::HTTPProxy` can raise, inherit from a common `Auth0::HTTPError`, and provide access to either the `RestClient::ExceptionWithResponse` or at least the `RestClient::Response` object in addition to the normal message.

joshcanhelp:
[...]

PR #170, opened 2019-06-15
Expose rate limiting information

philomory:
### Changes

This PR adds a new exception class, Auth0::RateLimitEncountered. In addition to the message and backtrace, this exception has attributes to hold the values of the [Auth0 Rate Limiting Headers](https://auth0.com/docs/policies/rate-limits#http-response-headers).

Thus, you can get access to this information when you encountered rate limiting, for example:
```ruby
try_again = true
begin
  auth0.clients
rescue Auth0::RateLimitEncountered => e
  if try_again
    try_again = false
    warn("Rate limiting encountered, waiting until #{Time.at(e.reset)} to try again")
    sleep(e.reset - Time.now.to_i)
    retry
  else
    raise e
  end
end
```

### References

This pull request address #158 

A future improvement might incorporate (optional) automatic retry-after-wait into the gem itself, but for now exposing the required information to library users is enough.

### Testing

This test adds unit tests, but not integration tests. 

* [ ] This change has been tested on the latest version of Ruby

### Checklist

### DIFF
--- a/lib/auth0/exception.rb
+++ b/lib/auth0/exception.rb
@@ -33,4 +33,12 @@ module Auth0
   class InvalidCredentials < Auth0::Exception; end
   # Invalid Auth0 API namespace
   class InvalidApiNamespace < Auth0::Exception; end
+  # Auth0 API rate-limiting encountered
+  class RateLimitEncountered < Auth0::Exception
+    attr_reader :limit, :remaining, :reset
+    def initialize(message,limit,remaining,reset)
+      super(message)
+      @limit, @remaining, @reset = limit, remaining, reset
+    end
+  end
 end
--- a/lib/auth0/exception.rb
+++ b/lib/auth0/exception.rb
@@ -52,7 +52,7 @@ module Auth0
   # Auth0 API rate-limiting encountered
   class RateLimitEncountered < Auth0::HTTPError
     def reset
-      Time.at(headers['X-RateLimit-Reset'])
+      Time.at(headers['X-RateLimit-Reset']).utc
     end
   end
 end
--- a/lib/auth0/exception.rb
+++ b/lib/auth0/exception.rb
@@ -50,7 +50,11 @@ module Auth0
   # Invalid Auth0 API namespace
   class InvalidApiNamespace < Auth0::Exception; end
   # Auth0 API rate-limiting encountered
-  class RateLimitEncountered < Auth0::HTTPError
+  # TODO: When making API-breaking changes, make this a subclass
+  # of Auth0::HTTPError directly rather than Auth0::Unsupported.
+  # It's currently under Unsupported to avoid breaking compatibility
+  # with prior gem versions that treated 429 errors as unknown errors.
+  class RateLimitEncountered < Auth0::Unsupported
     def reset
       Time.at(headers['X-RateLimit-Reset']).utc
     end
--- a/lib/auth0/exception.rb
+++ b/lib/auth0/exception.rb
@@ -2,19 +2,35 @@ module Auth0
   # Default exception in namespace of Auth0
   # if you want to catch all exceptions, then you should use this one.
   # Network exceptions are not included
-  class Exception < StandardError; end
+  class Exception < StandardError
+    attr_reader :error_data
+    def initialize(message,error_data={})
+      super(message)
+      @error_data = error_data
+    end
+  end
+  # Parent for all exceptions that arise out of HTTP error responses.
+  class HTTPError < Auth0::Exception
+    def headers
+      error_data[:headers]
+    end
+
+    def http_code
+      error_data[:code]
+    end
+  end
   # exception for unauthorized requests, if you see it,
   # probably Bearer Token is not set correctly
-  class Unauthorized < Auth0::Exception; end
+  class Unauthorized < Auth0::HTTPError; end
   # exception for not found resource, you query for an
   # unexistent resource, or wrong path
-  class NotFound < Auth0::Exception; end
+  class NotFound < Auth0::HTTPError; end
   # exception for unknown error
-  class Unsupported < Auth0::Exception; end
+  class Unsupported < Auth0::HTTPError; end
   # exception for server error
-  class ServerError < Auth0::Exception; end
+  class ServerError < Auth0::HTTPError; end
   # exception for incorrect request, you've sent wrong params
-  class BadRequest < Auth0::Exception; end
+  class BadRequest < Auth0::HTTPError; end
   # exception for timeouts
   class RequestTimeout < Auth0::Exception; end
   # exception for unset user_id, this might cause removal of
@@ -25,7 +41,7 @@ module Auth0
   # exception for an unset parameter
   class MissingParameter < Auth0::Exception; end
   # Api v2 access denied
-  class AccessDenied < Auth0::Exception; end
+  class AccessDenied < Auth0::HTTPError; end
   # Invalid parameter passed, e.g. empty where ID is required
   class InvalidParameter < Auth0::Exception; end
   # Invalid Auth0 credentials either client_id/secret for API v1
@@ -34,11 +50,9 @@ module Auth0
   # Invalid Auth0 API namespace
   class InvalidApiNamespace < Auth0::Exception; end
   # Auth0 API rate-limiting encountered
-  class RateLimitEncountered < Auth0::Exception
-    attr_reader :limit, :remaining, :reset
-    def initialize(message,limit,remaining,reset)
-      super(message)
-      @limit, @remaining, @reset = limit, remaining, reset
+  class RateLimitEncountered < Auth0::HTTPError
+    def reset
+      Time.at(headers['X-RateLimit-Reset'])
     end
   end
 end
--- a/spec/lib/auth0/mixins/httpproxy_spec.rb
+++ b/spec/lib/auth0/mixins/httpproxy_spec.rb
@@ -89,6 +89,27 @@ describe Auth0::Mixins::HTTPProxy do
           .and_raise(@exception)
         expect { @instance.send(http_method, '/test') }.to raise_error(Auth0::AccessDenied)
       end
+
+      it "should raise Auth0::RateLimitEncountered on send http #{http_method} method
+        to path defined through HTTP when 429 recieved" do
+        headers = {
+          'X-RateLimit-Limit'     => 10,
+          'X-RateLimit-Remaining' => 0,
+          'X-RateLimit-Reset'     => 1560564149
+        }
+        @exception.response = StubResponse.new({}, false, 429, headers)
+        allow(RestClient::Request).to receive(:execute).with(method: http_method,
+                                                             url: '/test',
+                                                             timeout: nil,
+                                                             headers: { params: {} },
+                                                             payload: nil)
+          .and_raise(@exception)
+        expect { @instance.send(http_method, '/test') }.to raise_error { |error|
+          expect(error).to be_a(Auth0::RateLimitEncountered)
+          expect(error).to have_attributes(limit: 10, remaining: 0, reset: 1560564149)
+        }
+      end      
+
       it "should raise Auth0::ServerError on send http #{http_method} method
         to path defined through HTTP when 500 received" do
         @exception.response = StubResponse.new({}, false, 500)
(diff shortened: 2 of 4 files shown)


damieng:
Hi @philomory thanks for this.  We have some internal discussions happening about how we want to handle and expose rate limiting across our various SDKs and want to make sure they are aligned. We'll let you know once we have a decision and if we need any tweaks to fit this PR in with that design once our discussions are concluded.

joshcanhelp:
@philomory - While I definitely want to make sure we do this correctly (as in, not make any decisions on this now that we have to back out of later), do you think it would make sense to pass the response headers to all the exceptions and allow for that additional param in `Auth0::Exception`?

```
module Auth0
    class Exception < StandardError
        attr_reader :headers
        def initialize(message, headers)
            super(message)
            @headers = headers
        end
    end
end
```

philomory:
That seems like a fairly sensible idea to me. I'm not sure entirely how much utility it has for most exception types, but it's certainly a more flexible implementation than one that is specific to the rate-limiting case, and thus probably more future-proof as well.

That said, if you are going to go that route, it may also be worth exposing the HTTP response code along with the headers; some of the exceptions pair 1-to-1 with a specific response code, but in the case of `Auth0::Unsupported`, there's currently no way to know what response code you received, since it's basically just "other". And if you're going to expose it on the `Auth0::Unsupported` exception, it probably makes sense to expose it on all of them for consistency.

The one caution in both of these cases is that it implies that all `Auth0::Exception` subclasses are going to always be about HTTP problems, and not, say, a failure of JWT verification (I know the library doesn't do JWT verification right now, but it might someday, I don't know). So if you add HTTP-related features to an exception base-class, it might be a good idea for it to be `Auth0::HTTPError` or `Auth0::TransportError` or something, which would then be the parent of all the current exceptions raised by `Auth0::Mixins::HttpProxy`, while itself being a subclass of `Auth0::Exception`.

At that point it _begins_ to feel like you should just expose the `RestClient::Response` object, but I actually don't like that plan, since it would make RestClient part of the gem's API (and I'm not super keen on RestClient anymore, myself).

Let me know how you'd like to proceed, and whether you want me to tweak my PR, or whether you want to close this PR and implement the changes yourselves.

joshcanhelp:
@philomory - Thank you for your thoughtful reply. As you mentioned, those exceptions are not always HTTP-related. Maybe that second parameter could be `additional information` or similar? In the case of JWT validation, say, that could be expected values that aren't in danger of being exposed. 

I like the idea of including the HTTP code as well. Could be just a generic `error_code`, which could include `error` responses from our API. Feels like that makes sense as the second parameter and the headers/other as the third. 

> At that point it begins to feel like you should just expose the `RestClient::Response` object

We'll be replacing RestClient in a future version, likely a major. I do like the idea of an `HttpError` parent class, though. 

Again, thank you for all this. We're working on a release for this week so if you're able to make these changes, I'd be happy to merge and get this out as part of that.

philomory:
So, a basic sketch of what you've outlined above might look like this:

```ruby
  class Exception < StandardError
    attr_reader :error_code, :additional_information
    def initialize(message,error_code=nil,additional_information={})
      super(message)
      @error_code, @additional_information = error_code, additional_information
    end
  end
  class HTTPError < Auth0::Exception; end
  class Unauthorized < Auth0::HTTPError; end
  ...
```

And get used like this:

```
          case result.code
          when 200...226 then safe_parse_json(result.body)
          when 400       then raise Auth0::BadRequest.new(result.body, result.code, headers: result.headers)
          when 401       then raise Auth0::Unauthorized.new(result.body, result.code, headers: result.headers)
          when 403       then raise Auth0::AccessDenied.new(result.body, result.code, headers: result.headers)
          when 404       then raise Auth0::NotFound.new(result.body, result.code, headers: result.headers)
          when 429       then raise Auth0::RateLimitEncountered.new(result.body, result.code, headers: result.headers)
          when 500       then raise Auth0::ServerError.new(result.body, result.code, headers: result.headers)
          else                raise Auth0::Unsupported.new(result.body, result.code, headers: result.headers)
          end
```
In this case the consumer of the exception could get access to the headers via something like:

```ruby
begin
  auth0.some_call(args)
rescue Auth0::RateLimitEncountered => e
  reset = e.additional_information[:headers]['X-RateLimit-Reset']
  wait_until(reset)
  retry
end
```

You could add a convenience method to `Auth0::HTTPError` that returned the headers, e.g. 
```ruby
class HTTPError < Auth::Exception
  def headers
    additional_information[:headers]
  end
end
```
... so it could be called more cleanly, like `e.headers['X-RateLimit-Reset']`, for users who know what kind of exception they are working with.

The only awkward thing about this design is that, for non-HTTP errors that have additional information to provide, but nothing analogous to an HTTP response code, code raising that exception has to explicitly pass `nil` as the second argument; you could use keyword arguments to clean that up, but it does get a little cumbersome with long keyword names like additional_information, e.g.
```ruby
raise Auth0::RateLimitEncountered.new(result.body, error_code: result.code, additional_information: { headers: result.headers })
```

Anyway, with the holiday in the US this week I'm not sure I'll have the time to put together an updated version of the PR this week, not if we also want corresponding unit tests and acceptance tests.

joshcanhelp:
@philomory - Thanks again for your work here. I totally forgot about the short week so next week is totally fine. I'm not sure if we'll be able to get the rest of the changes in place before then on our end either so no rush. I'd rather wait and get this in. 

Everything is looking great here. I think the long var name and missing code make a case for just using a single var for all that data we're getting back, like `error_data`:

```ruby
class Exception < StandardError
  attr_reader :error_data
  def initialize(message,error_data={})
    super(message)
    @error_data = error_data
  end
end

# ...

case result.code
when 200...226 then safe_parse_json result.body
when 400 then Auth0::BadRequest.new result.body, { http_code: result.code, headers: result.headers }

# ...

class HTTPError < Auth::Exception
  def headers
    error_data[:headers]
  end

  def http_code
    error_data[:http_code]
  end
end
```

philomory:
I've pushed a patch which adds the changes we discussed in terms of an `error_data` attribute on `Auth0::Exception`, `headers` and `http_code` convenience methods on `Auth0::HTTPError`, and a `reset` convenience-method on `Auth0::RateLimitEncountered` that basically returns `Time.at(headers['X-RateLimit-Reset']).utc`.

I also tweaked the new unit tests I had already added, so that they would work with the new setup. I did _not_ add any additional unit tests for the general changes to `Auth0::Exception` or `Auth0::HTTPError`, only for the new `Auth0::RateLimitEncountered` error - honestly, I wasn't sure where those sorts of tests would go or what they would look like. I also didn't add any integration tests.

joshcanhelp:
@philomory - Thanks once again! This looks great at first glance. I'm going to take a closer look Monday and get back to you with feedback, if I have any. The tests you've already added look fine, I don't know that integration tests for internals like this would do any more good than the stubbed ones.

spec/lib/auth0/mixins/httpproxy_spec.rb
```diff
@@ -148,6 +169,26 @@
           .and_raise(@exception)
         expect { @instance.send(http_method, '/test') }.to raise_error(Auth0::Unauthorized)
       end
+      
+      it "should raise Auth0::RateLimitEncountered on send http #{http_method} method
```
joshcanhelp: I'm struggling to understand what this test is supposed to be doing. I see that `headers` are different between the two. The expects are the same though. Can you explain?

spec/lib/auth0/mixins/httpproxy_spec.rb
```diff
@@ -148,6 +169,26 @@
...
+        @exception.response = StubResponse.new({}, false, 429,headers)
+        allow(RestClient::Request).to receive(:execute).with(method: http_method,
+                                                             url: '/test',
+                                                             timeout: nil,
+                                                             headers: nil,
+                                                             payload: '{}')
+          .and_raise(@exception)
+        expect { @instance.send(http_method, '/test') }.to raise_error { |error|
+          expect(error).to be_a(Auth0::RateLimitEncountered)
+          expect(error).to have_attributes(reset: Time.at(1560564149))
```
joshcanhelp: Maybe check that we also have the headers and the code? Either here or in another test.

lbalmaceda:
@joshcanhelp @philomory is there anything I can do to unblock this one? Feels like a good, needed contribution to this SDK. Happy to resolve any doubts.

spec/lib/auth0/mixins/httpproxy_spec.rb
philomory: These are essentially the same, but one is in the block that tests the `GET`/`DELETE` method, and one is in the block that tests the `POST`/`PUT`/`PATCH` methods; I was simply following the pattern established by the other, existing tests.

spec/lib/auth0/mixins/httpproxy_spec.rb
philomory: This is what I was referring to when I said I wasn't sure where to put tests that were general to `Auth0::Exception` or `Auth0::HTTPError`; the `code` and `headers` features are not specific to `Auth0::RateLimitEncountered` (to say nothing of the `error_data` feature), so I wasn't sure it made sense to test them in that context; that said, I can certainly add tests for those features there if desired.

philomory: These tests have been added now.

joshcanhelp:
@philomory - Thanks for the update here! I'm out of office for a bit but I'll pull this down and run through it when I'm back to get it in the next release.

philomory:
Apropos of very little, what does the `CH: Added` label mean?

joshcanhelp:
@philomory - Indicates a new feature of some kind. We use that label to generate the changelog.

I'm pulling this down and testing today/tomorrow and will get this in a release this week 👍

lib/auth0/exception.rb
```diff
@@ -25,12 +41,18 @@ class MissingClientId < Auth0::Exception; end
...
+  class AccessDenied < Auth0::HTTPError; end
   # Invalid parameter passed, e.g. empty where ID is required
   class InvalidParameter < Auth0::Exception; end
   # Invalid Auth0 credentials either client_id/secret for API v1
   # or JWT for API v2/
   class InvalidCredentials < Auth0::Exception; end
   # Invalid Auth0 API namespace
   class InvalidApiNamespace < Auth0::Exception; end
+  # Auth0 API rate-limiting encountered
+  class RateLimitEncountered < Auth0::HTTPError
```
joshcanhelp: Should we extend `Auth0::Unsupported` here instead of `Auth0::HTTPError`? Without that, we're not backwards-compatible (we're now throwing a different exception and, for apps that are rescuing `Auth0::Unsupported` in the case of a rate limit hit, that exception will bubble up).

lib/auth0/exception.rb
```diff
@@ -25,12 +41,18 @@ class MissingClientId < Auth0::Exception; end
...
+    def reset
+      Time.at(headers['X-RateLimit-Reset']).utc
+    end
+  end
```
joshcanhelp: Do you think it would be helpful to have a `limit` method here as well?

lib/auth0/exception.rb
philomory: Ah, hm, interesting point. From an API-compatibility point, yes; from a semantic standpoint, no?

Actually, I wonder if `Auth0::Unsupported` should just become an alias for `Auth0::HTTPError`. After all, what _is_ `Auth0::Unsupported` if not "An HTTP error that this gem doesn't have any more specific classification for"?

E.g., instead of making `Auth0::RateLimitEncountered` inherit from `Auth0::Unsupported`, just have this line in there:

`Unsupported = Auth0::HTTPError`

philomory: Hm. Actually, maybe that's not great either, since it would broaden the scope of `Auth0::Unsupported`; `rescue Auth0::Unsupported` would not rescue error types that it didn't previously.

lib/auth0/exception.rb
philomory: I couldn't think of a use-case for it, given that you won't find out until after you've already hit the limit; but it would be pretty easy to include, if we want it.

lib/auth0/exception.rb
joshcanhelp: I'm more worried about breaking folks on a minor than semantics, at this point. We're due for a major on this SDK (a number of deprecated methods and addressing the HTTP client we're using) and could remove that extension then.

philomory: Ok, I've made `Auth0::RateLimitEncountered` a subclass of `Auth0::Unsupported`, with a comment explaining the reason and a TODO item to move it at a major version change.

philomory:
The snyk CI check is now failing, although from what I can tell (I'm not very familiar with snyk) it seems to be unrelated to this PR - if I'm reading correctly, a vulnerability apparently exists in YARD, which this gem has as a development dependency.

joshcanhelp:
@philomory - Deps fixed in `master`. Just rebase your base and push and we should be good to go here!

philomory:
All checks now pass 😄

PR #169, opened 2019-06-11
Add missing User Management API endpoints

joshcanhelp:
### Changes

Adds all missing User endpoints:

- `Auth0::Api::V2::Users.get_roles`
- `Auth0::Api::V2::Users.remove_roles`
- `Auth0::Api::V2::Users.add_roles`
- `Auth0::Api::V2::Users.get_enrollments`
- `Auth0::Api::V2::Users.get_permissions`
- `Auth0::Api::V2::Users.remove_permissions`
- `Auth0::Api::V2::Users.add_permissions`
- `Auth0::Api::V2::Users.generate_recovery_code`
- `Auth0::Api::V2::Users.invalidate_browsers`

### Testing

Will add integration tests in another PR after Roles

* [x] This change adds unit test coverage
* [x] This change has been tested on the latest version of Ruby (2.6.0)

### Checklist

* [x] All existing and new tests complete without errors
* [x] Rubocop passes on all added/modified files
* [x] All active GitHub checks have passed

PR #174, opened 2019-06-27
Add Roles and Users endpoints integration tests

joshcanhelp:
### Changes

- Add integration tests (recorded with VCR) for new Roles and Users endpoints.
- Rename `unlink_users_account` to `unlink_user_account`, added alias and test
- Add `delete_with_body` method to send a body rather than params with a `DELETE` request

**Note for reviewers:**

- New VCR recordings were add in `spec/fixtures/vcr_cassettes/Auth0_Api_V2_Users/` and `spec/fixtures/vcr_cassettes/Auth0_Api_V2_Roles/`. Sensitive information (API token) is automatically filtered out. 
- The other YML recording files have either a changed format, changed date, or removed dummy API token (the one that was there is not valid but makes more sense for them all to have `API_TOKEN`). 
- In general, YML files can be ignored. Without those, review is ~360 changed lines total. 

### References

- User methods added in #169 (adjust naming in #173 and #175)
- Role methods added in #172 (adjust naming in #173)

### Testing

* [ ] This change adds unit test coverage (added in original PR)
* [x] This change has been tested on the latest version of Ruby (2.6.0)


lib/auth0/api/v2/users.rb
```diff
@@ -220,19 +221,19 @@ def remove_user_roles(user_id, roles)
           raise Auth0::MissingUserId, 'Must supply a valid user_id' if user_id.to_s.empty?
           validate_strings_array roles
           path = "#{users_path}/#{user_id}/roles"
-          delete(path, { roles: roles })
+          delete_with_body path, roles: roles
```
lbalmaceda: I see inconsistencies on the syntax when calling a function with parameters.

`delete(path, {param: value}` vs `delete path, param:value`. 

Should we add a linter as a precommit for this repo as well or is this change on purpose?

lib/auth0/mixins/httpproxy.rb
```diff
@@ -20,6 +20,8 @@ module HTTPProxy
                      call(:get, url(safe_path), timeout, get_headers)
                    elsif method == :delete
                      call(:delete, url(safe_path), timeout, add_headers({params: body}))
```
lbalmaceda: what is this `params:body` header for? I guess you meant to send a query instead?

joshcanhelp: Oddly, sending it as `params:` converts it to a query ... one of the many quirks of this HTTP library

lib/auth0/api/v2/users.rb
joshcanhelp: There is a linter that I run, `rubocop`, but it doesn't pick this up

PR #179, opened 2019-07-11
Add Management API Anomaly endpoints

makoto-matsumoto:
### Changes

Add endpoint for Anomaly API

### References

https://auth0.com/docs/api/management/v2#!/Anomaly/get_ips_by_id
https://auth0.com/docs/api/management/v2#!/Anomaly/delete_ips_by_id

### Testing

### Checklist


spec/integration/lib/auth0/api/v2/api_anomaly_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'spec_helper'
+describe Auth0::Api::V2::Anomaly do
+  let(:client) { Auth0Client.new(v2_creds) }
+  let(:ip) { '192.0.2.0' }
+
+  describe '.check_if_ip_is_blocked', vcr: true do
+    it 'should rerurn 200 response code' do
```
joshcanhelp:
```suggestion
    it 'should return 200 response code' do
```

lib/auth0/api/v2/anomaly.rb
```diff
@@ -0,0 +1,36 @@
+module Auth0
+  module Api
+    module V2
+      # Methods to use the anomaly endpoints
+      module Anomaly
+        # Use this route to determine if a given ip is currently blocked
```
joshcanhelp:
```suggestion
        # Use this route to determine if a given IP is currently blocked
```

spec/fixtures/vcr_cassettes/Auth0_Api_V2_Anomaly/_check_if_ip_is_blocked/should_rerurn_200_response_code.yml
```diff
@@ -0,0 +1,65 @@
...
+      Accept:
+      - "*/*"
+      Accept-Encoding:
+      - gzip, deflate
+      User-Agent:
+      - rest-client/2.0.2 (darwin18.6.0 x86_64) ruby/2.6.3p62
+      Content-Type:
+      - application/json
+      Auth0-Client:
+      - eyJuYW1lIjoicnVieS1hdXRoMCIsInZlcnNpb24iOiI0LjUuMCJ9
```
joshcanhelp: Your attention to detail is commendable!

I'll make a note to anonymize this in our existing cassettes. Nothing secret here but probably best not to have specific application data in here.

lib/auth0/api/v2/anomaly.rb
```diff
@@ -0,0 +1,36 @@
...
+        # by the failed login to multiple user accounts trigger.
+        # @see https://auth0.com/docs/api/management/v2#!/Anomaly/get_ips_by_id
+        # @param ip [string] The IP to check.
+        def check_if_ip_is_blocked(ip)
+          raise Auth0::InvalidParameter, 'Must specify an IP' if ip.to_s.empty?
```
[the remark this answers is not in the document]
makoto-matsumoto: I think that it is better to return the original response of Auth0 as much as possible.
If the format is incorrect, Auth0 will return:

```
RESPONSE CODE : 400
RESPONSE BODY
{
  "statusCode": 400,
  "error": "Bad Request",
  "message": "Path validation error: 'Object didn't pass validation for format ipv4: 9919191' on property id (The ip to check).",
  "errorCode": "invalid_uri"
}
```

spec/fixtures/vcr_cassettes/Auth0_Api_V2_Anomaly/_check_if_ip_is_blocked/should_rerurn_200_response_code.yml
makoto-matsumoto: Thank you for your advice!  
However, this value has been updated to the same value as the test already included in the master.
Ex:
[should_add_a_token_to_the_blacklist.yml:19](https://github.com/auth0/ruby-auth0/blob/master/spec/fixtures/vcr_cassettes/Auth0_Api_V2_Blacklists/_add_token_to_blacklist/should_add_a_token_to_the_blacklist.yml#L19)

You might want to include client id in filter_sensitive_data as well.

spec/spec_helper.rb
``` 
config.filter_sensitive_data('CLIENT_SECRET') { ENV['CLIENT_SECRET'] }
config.filter_sensitive_data('API_TOKEN') { ENV['MASTER_JWT'] }
```

joshcanhelp: I think that's probably the best way to deal with it. My main reason for commenting was that I appreciate you updating to the same value as the rest 👍

lib/auth0/api/v2/anomaly.rb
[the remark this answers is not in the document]
joshcanhelp: The main thing I was thinking here is that we could save an HTTP call if we can catch the error (like we do in other cases). I'll check with the team to see what they think.

PR #209, opened 2020-03-30
Pass client_id, audience at Auth0::Api::V2::ClientGrants#client_grants

hkdnet:
### Changes

I'd like to fetch a client grant by its client id. 
We can't call `Auth0::Api::V2::ClientGrants#client_grants` with `client_id` while `GET /api/v2/client-grants` accepts `client_id`.

### References

Public doc for `GET /api/v2/client-grants`
https://auth0.com/docs/api/management/v2#!/client_grants/get_client_grants

### Testing

    - I'm very happy with adding integration tests for this but I can't run `master`'s integration tests on my local.

### Checklist

    * No. But `master` also fails.

### DIFF
--- a/lib/auth0/api/v2/client_grants.rb
+++ b/lib/auth0/api/v2/client_grants.rb
@@ -7,11 +7,13 @@ module Auth0
 
         # Retrieves a list of all client grants.
         # @see https://auth0.com/docs/api/management/v2#!/client_grants/get_client_grants
+        # @param client_id [string] The client_id of the client grant to retrieve.
         # @param page [int] Page number to get, 0-based.
         # @param per_page [int] Results per page if also passing a page number.
         # @return [json] Returns the client grants.
-        def client_grants (page: nil, per_page: nil)
+        def client_grants (client_id: nil, page: nil, per_page: nil)
           request_params = {
+            client_id: client_id,
             page: page,
             per_page: per_page
           }
--- a/lib/auth0/api/v2/client_grants.rb
+++ b/lib/auth0/api/v2/client_grants.rb
@@ -8,12 +8,14 @@ module Auth0
         # Retrieves a list of all client grants.
         # @see https://auth0.com/docs/api/management/v2#!/client_grants/get_client_grants
         # @param client_id [string] The client_id of the client grant to retrieve.
+        # @param audience [string] The audience of the client grant to retrieve.
         # @param page [int] Page number to get, 0-based.
         # @param per_page [int] Results per page if also passing a page number.
         # @return [json] Returns the client grants.
-        def client_grants (client_id: nil, page: nil, per_page: nil)
+        def client_grants (client_id: nil, audience: nil, page: nil, per_page: nil)
           request_params = {
             client_id: client_id,
+            audience: audience,
             page: page,
             per_page: per_page
           }
--- a/spec/lib/auth0/api/v2/client_grants_spec.rb
+++ b/spec/lib/auth0/api/v2/client_grants_spec.rb
@@ -14,6 +14,7 @@ describe Auth0::Api::V2::ClientGrants do
       expect(@instance).to receive(:get).with(
         '/api/v2/client-grants',
         client_id: nil,
+        audience: nil,
         page: nil,
         per_page: nil
       )
@@ -24,6 +25,7 @@ describe Auth0::Api::V2::ClientGrants do
       expect(@instance).to receive(:get).with(
         '/api/v2/client-grants',
         client_id: nil,
+        audience: nil,
         page: 1,
         per_page: 2
       )
--- a/spec/lib/auth0/api/v2/client_grants_spec.rb
+++ b/spec/lib/auth0/api/v2/client_grants_spec.rb
@@ -21,6 +21,19 @@ describe Auth0::Api::V2::ClientGrants do
       expect { @instance.client_grants }.not_to raise_error
     end
 
+    it 'is expected to send get /api/v2/client-grants/ with client_id and audience' do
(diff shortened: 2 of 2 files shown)


hkdnet:
I met two GitHub check failures.

## snyk

GitHub check `ci/circleci: snyk` is marked as Required but I think this is not required for all pull requests.
CircleCI workflow snyk only runs with context `snyk-env`, which is not available for me.
So snyk workflow for my PR is skipped as `UNAUTHORIZED`.
I believe this is a configuration error. Can someone take a look? 

## codecov/changes

Checked the error and read https://docs.codecov.io/docs/unexpected-coverage-changes but I have no idea why this happens.
Let me investigate more :pray:

lib/auth0/api/v2/client_grants.rb
```diff
@@ -7,11 +7,13 @@ module ClientGrants
 
         # Retrieves a list of all client grants.
         # @see https://auth0.com/docs/api/management/v2#!/client_grants/get_client_grants
+        # @param client_id [string] The client_id of the client grant to retrieve.
         # @param page [int] Page number to get, 0-based.
         # @param per_page [int] Results per page if also passing a page number.
         # @return [json] Returns the client grants.
-        def client_grants (page: nil, per_page: nil)
+        def client_grants (client_id: nil, page: nil, per_page: nil)
```
lbalmaceda, line 14: This would be less breaking if put at the end. WDYT @davidpatrick ?

Also, I think we should include `audience` as well. Also optional @hkdnet  👍

spec/lib/auth0/api/v2/client_grants_spec.rb
```diff
@@ -22,10 +23,11 @@
     it 'is expected to send get /api/v2/client-grants/ with pagination' do
       expect(@instance).to receive(:get).with(
         '/api/v2/client-grants',
+        client_id: '1',
         page: 1,
         per_page: 2
       )
-      expect { @instance.client_grants(page: 1, per_page: 2) }.not_to raise_error
```
lbalmaceda, line 28: keep this test, and add a new one for each introduced property. This way you fix the "Codecov: Changes" failing status check.

lbalmaceda:
Thanks for pointing out the failing snyk check. I've updated that on this PR https://github.com/auth0/ruby-auth0/pull/210.

spec/lib/auth0/api/v2/client_grants_spec.rb
hkdnet: Sure. Thank you!

lib/auth0/api/v2/client_grants.rb
hkdnet: > This would be less breaking if put at the end.

An order of keyword arguments does not matter in Ruby.
I prefer putting before `page` and `per_page` because giving `client_id` changes responses more than other parameters. (If you prefer another style, I'm happy to follow it!)

hkdnet: Added `audience` as well. f7c3c9d

lbalmaceda: Those calling it with named arguments will not have a problem 👍 however those still using the standard way of passing arguments would run into a breaking change, as the argument they were sending is now pointing to a different variable. 

Say someone today has this call:

```ruby
# This works today
client_grants(1, 2) 
```

That would be the same as calling with named arguments: 
```ruby
# Without this PR
# def client_grants (page: nil, per_page: nil)
client_grants(page = 1, per_page = 2) 
```

Now say we put new parameters at the beginning:

```ruby
# With this PR
# def client_grants (audience = nil, client_id = nil, page: nil, per_page: nil)
client_grants(audience = 'myaudience', client_id = 'yourclient', page = 1, per_page = 2)
```

That works, but as soon as we remove the name of the params, the user who had working code is now running into issues:

```ruby
# With this PR
# def client_grants (audience = nil, client_id = nil, page: nil, per_page: nil)
client_grants(1, 2)
```
Behavior changes as `1` is taken as the audience, and `2` as the client id.

AFAIK Ruby works that way. I could be wrong, it's been years without coding ruby. 😅

hkdnet: Ruby has ordinary parameters, optional parameters, and keyword parameters.
Your concern is correct with optional parameters but this method is defined with keyword parameters.

Labels are required to pass values as keyword arguments. All users pass values with labels (or pass no values).

```ruby
client.client_grants(1, 2) # this does not work. ArugmentError is raised
client.client_grants(page: 1, per_page: 2) # works
client.client_grants() # This also works because both page and per_page are optional.
```

In this case, adding keyword parameters before the existing keywords does not break anything. Here is a smaller example.

```ruby
def f(a: nil)
  puts "a = #{a}"
end

f(a: 1) # a = 1

def f(b: nil, a: nil)
  puts "a = #{a}, b = #{b}"
end

f(a: 1, b: 2) # a = 1, b = 2
f(b: 2, a: 1) # a = 1, b = 2

# This still works well :)
f(a: 1) # a = 1, b = 
```

You can confirm users can use `client_grants` the same as before with these examples. These specs work well without changing.

 - https://github.com/auth0/ruby-auth0/pull/209/files#diff-19c1e1d1d6e6aa44a0c115a794fa4a34R21
- https://github.com/auth0/ruby-auth0/pull/209/files#diff-19c1e1d1d6e6aa44a0c115a794fa4a34R45

lbalmaceda: Perfect. Then this is solved 👍 Thanks for your contribution

hkdnet:
Thank you @lbalmaceda !

PR #222, opened 2020-06-02
Improve OIDC compliance: add signature validation support [SDK-987]

Widcket:
### Changes

This update improves the SDK support for **OpenID Connect (OIDC)**. In particular, it adds support for **signature validation**. Additional PRs will further expand the SDK support for OIDC.

#### What’s being added in this PR

- **validate_id_token method**
`validate_id_token(id_token, algorithm: nil, leeway: 60, nonce: nil, max_age: nil, issuer: nil, audience: nil)`
Public method in `Auth0::Client` that validates its parameters, creates a new instance of `IdTokenValidator` and calls it with the passed parameters.
Plus tests.

- **IdTokenValidator class**
Class that contains the logic to validate the signature and claims of the ID token. The signature validation logic was implemented in this PR; the claims validation logic will be implemented in the next PR.
Plus tests.

- **HS256 class**
`Auth0::Algorithm::HS256`
Class that represents the `HS256` algorithm and holds the shared secret. 
Plus tests.

- **RS256 class**
`Auth0::Algorithm::RS256`
Class that represents the `RS256` algorithm and holds the JWK set url. Also contains the logic to fetch the JWK set from that url and cache it in an in-memory cache with a configurable TTL.
Plus tests.

### Testing

### Checklist


lib/auth0/api/authentication_endpoints.rb
```diff
@@ -502,6 +506,51 @@ def unlink_user(access_token, user_id)
...
+        end
+
+        context = {
+          issuer: issuer || "https://#{@domain}/",
+          audience: audience || @client_id,
+          algorithm: algorithm || Auth0::Algorithm::RS256.jwks_url("https://#{@domain}/.well-known/jwks.json"),
+          leeway: leeway
+        }
+
+        context[:nonce] = nonce unless nonce.nil?
```
lbalmaceda, line 546: it could be added anyway above in the { } definition and checked down in the claims validation section. There's no escape from that later check, so it seems best to simplify these two here.

lib/auth0/mixins/validation.rb
```diff
@@ -20,6 +28,118 @@ def validate_permissions_array(permissions)
...
+          rescue JWT::IncorrectAlgorithm
+            alg = decoded_id_token[:header]['alg']
+            raise Auth0::InvalidIdToken, "Signature algorithm of \"#{alg}\" is not supported. Expected the ID token to be signed with \"#{algorithm_name}\""
+          rescue JWT::DecodeError
+            kid = decoded_id_token[:header]['kid']
+            raise Auth0::InvalidIdToken, "Could not find a public key for Key ID (kid) \"#{kid}\""
+          end
+        end
+
+        def validate_claims(decoded_id_token); end
```
lbalmaceda, line 76: add an inline comment `//TODO` to signal this will be added on a separate PR

lib/auth0/mixins/validation.rb
```diff
@@ -20,6 +28,118 @@ def validate_permissions_array(permissions)
...
+        # @see https://auth0.com/docs/tokens/concepts/signing-algorithms
+        class RS256
+          include Auth0::Mixins::HTTPProxy
+
+          @@cache = Zache.new.freeze
+
+          class << self
+            private :new
+
+            # Create a new instance passing the shared secret.
```
lbalmaceda, line 113: these docs are for the HS256 case

lib/auth0/mixins/validation.rb
```diff
@@ -20,6 +28,118 @@ def validate_permissions_array(permissions)
...
+          end
+
+          # Fetches the JWK set from the in-memory cache, or from the url as a fallback.
+          # @return [hash] A JWK set.
+          def jwks
+            previous_value = @@cache.last(:jwks)[:value] unless @@cache.last(:jwks).nil?
+
+            @@cache.get(:jwks, lifetime: @lifetime, dirty: true) do
+              result = get(@jwks_url)
+              result.is_a?(Hash) && result.key?('keys') ? result : previous_value
```
lbalmaceda, line 138: I'd expect an explicit set but maybe the library is doing that for you. Anyway, will the cache be updated with this result for the next call?

spec/lib/auth0/mixins/validation_spec.rb
```diff
@@ -0,0 +1,207 @@
...
+    end
+
+    it 'is not expected to fetch the jwks from the web when there is a value cached' do
+      stub_jwks
+      instance = Auth0::Algorithm::RS256.jwks_url(JWKS_URL)
+      instance.jwks
+      instance.jwks
+
+      WebMock.after_request do
+        expect(a_request(:get, JWKS_URL)).to have_been_made.at_most_once
```
lbalmaceda, line 198: let's be more explicit here, only one request should come through 

```suggestion
        expect(a_request(:get, JWKS_URL)).to have_been_made.once
```

lib/auth0/mixins/validation.rb
```diff
@@ -20,6 +28,118 @@ def validate_permissions_array(permissions)
...
+            raise Auth0::InvalidParameter, 'Must supply a valid lifetime' unless lifetime.is_a?(Integer) && lifetime >= 0
+
+            @lifetime = lifetime
+            @jwks_url = jwks_url
+          end
+
+          # Fetches the JWK set from the in-memory cache, or from the url as a fallback.
+          # @return [hash] A JWK set.
+          def jwks
+            previous_value = @@cache.last(:jwks)[:value] unless @@cache.last(:jwks).nil?
```
lbalmaceda, line 134: what is this for, why do you need to invoke `last` on the cache? and what's the value of `:jwks` and `:value`?

lib/auth0/api/authentication_endpoints.rb
```diff
@@ -502,6 +506,51 @@ def unlink_user(access_token, user_id)
...
+        end
+
+        unless algorithm.nil? || algorithm.is_a?(Auth0::Algorithm::RS256) || algorithm.is_a?(Auth0::Algorithm::HS256)
+          raise Auth0::InvalidParameter, "Signature algorithm of \"#{algorithm}\" is not supported"
+        end
+
+        context = {
+          issuer: issuer || "https://#{@domain}/",
+          audience: audience || @client_id,
+          algorithm: algorithm || Auth0::Algorithm::RS256.jwks_url("https://#{@domain}/.well-known/jwks.json"),
```
lbalmaceda, line 542: if I get this right, a new context and a new algorithm are set every time the `validate_id_token` method is called. If so, how is the JWKS cache being kept for the next time I want to validate a token?

A way to assert this would be spying the request class, calling `validate_id_token` twice with the same default config you're setting here (not passing an algorithm in the arguments), and checking if the URL was called once, as expected.

lib/auth0/mixins/validation.rb
```diff
@@ -20,6 +28,118 @@ def validate_permissions_array(permissions)
...
+        end
+
+        def validate(id_token)
+          begin
+            result = JWT.decode id_token, nil, false, {}
+          rescue JWT::DecodeError
+            raise Auth0::InvalidIdToken, 'ID token could not be decoded'
+          end
+
+          decoded_id_token = { payload: result.first, header: result.last, string: id_token }
```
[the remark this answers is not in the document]
Widcket: Yes, the [jwt](https://github.com/jwt/ruby-jwt) library performs the signature validation via the `decode` method, and to call `decode` I need the complete JWT string. There's no exposed API just for the signature validation AFAIK.

lib/auth0/mixins/validation.rb
```diff
@@ -20,6 +28,118 @@ def validate_permissions_array(permissions)
...
+          validate_signature decoded_id_token
+          validate_claims decoded_id_token
+        end
+
+        private
+
+        def validate_signature(decoded_id_token)
+          algorithm = @context[:algorithm]
+          algorithm_name = algorithm.class.to_s.split('::').last
```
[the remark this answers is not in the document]
Widcket: Good call, I'll change it.

lib/auth0/api/authentication_endpoints.rb
```diff
@@ -502,6 +506,51 @@ def unlink_user(access_token, user_id)
...
+        raise Auth0::InvalidParameter, 'Must supply a valid leeway' unless leeway.is_a?(Integer) && leeway >= 0
+        raise Auth0::InvalidParameter, 'Must supply a valid nonce' unless nonce.nil? || !nonce.to_s.empty?
+        raise Auth0::InvalidParameter, 'Must supply a valid issuer' unless issuer.nil? || !issuer.to_s.empty?
+        raise Auth0::InvalidParameter, 'Must supply a valid audience' unless audience.nil? || !audience.to_s.empty?
+
+        unless max_age.nil? || (max_age.is_a?(Integer) && max_age >= 0)
+          raise Auth0::InvalidParameter, 'Must supply a valid max_age'
+        end
+
+        unless algorithm.nil? || algorithm.is_a?(Auth0::Algorithm::RS256) || algorithm.is_a?(Auth0::Algorithm::HS256)
```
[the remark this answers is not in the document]
Widcket: Good idea, thanks for pointing it out!

lib/auth0/mixins/validation.rb
Widcket: That was added as a logical safety measure to make sure I do not end up caching a `nil` value, in the event that the network request fails in a way that does not throw an exception (e.g an unexpected response with a successful status code, like sometimes happens in an outage). In that case, the cache will keep the last known value for another `lifetime` period.
I added the `last` method to the cache class to be able to just fetch the last cached value, expired or not.
`:jwks` is the cache key for the jwks. `:value` is the key to access the actual jwks content in the cache entry, that is a hash containing other [values](https://github.com/yegor256/zache/blob/master/lib/zache.rb#L193) (the start time and the lifetime).

lib/auth0/mixins/validation.rb
Widcket: Ruby has implicit returns, so that line (the last one in the block) will return the jwks obtained from the network call or the last known value otherwise.

Widcket: The [return value](https://github.com/yegor256/zache/blob/master/lib/zache.rb#L187) of that block is what ends up in the cache.

lib/auth0/mixins/validation.rb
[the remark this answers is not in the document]
lbalmaceda: Ok, then we need to refactor this because as of today it is decoding the same token twice. 

Rename the `validate_signature` to something like `validate_signature_and_decode` and make it return the payload claims dictionary/map that the `validate_claims` method would receive. As soon as the algorithm name is verified along with the signature, the header claims are no longer required.

lib/auth0/api/authentication_endpoints.rb
Widcket: You're right, I'll change it.

lib/auth0/api/authentication_endpoints.rb
Widcket: The JWKS cache is a private static variable, that additionally is frozen to prevent it from being reassigned. So there is always a single instance of the cache per Ruby process.

Widcket: There are already cache tests that verify that, or are you referencing another possible test case?

lib/auth0/mixins/validation.rb
lbalmaceda: That's good thanks for clarifying. Remember the cache miss scenario I mentioned in the general review comment.

lib/auth0/api/authentication_endpoints.rb
```diff
@@ -502,6 +506,51 @@ def unlink_user(access_token, user_id)
...
+      # @param max_age [integer] The max_age value sent during authentication.
+      #   Must be a non-negative value.
+      # @param issuer [string] The expected issuer claim value.
+      #   Defaults to "https://YOUR_AUTH0_DOMAIN/".
+      # @param audience [string] The expected audience claim value.
+      #   Defaults to your Auth0 Client ID.
+      # rubocop:disable Metrics/MethodLength, Metrics/AbcSize, Metrics/ParameterLists
+      def validate_id_token(id_token, algorithm: nil, leeway: 60, nonce: nil, max_age: nil, issuer: nil, audience: nil)
+        raise Auth0::InvalidParameter, 'ID token is required but missing' if id_token.to_s.empty?
+        raise Auth0::InvalidParameter, 'Must supply a valid leeway' unless leeway.is_a?(Integer) && leeway >= 0
```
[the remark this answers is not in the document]
Widcket: That kind of check is [used](https://github.com/auth0/ruby-auth0/blob/master/lib/auth0/api/authentication_endpoints.rb#L128) [extensively](https://github.com/auth0/ruby-auth0/blob/master/lib/auth0/api/authentication_endpoints.rb#L104) in the SDK codebase to check that a string param is not `nil` or empty in a single check. `nil.to_s` equals `''`, so it all boils down to the string not being empty.

lib/auth0/mixins/validation.rb
Widcket: Though `:value` should be in the definition of `last` though. I'll change it.

Widcket: Chaged in efdf3feec312ed9d2e8a91fa3ea9b625373033ec

lib/auth0/mixins/validation.rb
Widcket: Fixed in efdf3feec312ed9d2e8a91fa3ea9b625373033ec

lib/auth0/mixins/validation.rb
Widcket: Added in efdf3feec312ed9d2e8a91fa3ea9b625373033ec

lib/auth0/mixins/validation.rb
```diff
@@ -20,6 +28,118 @@ def validate_permissions_array(permissions)
...
+          options = { algorithms: [algorithm_name], verify_expiration: false, verify_not_before: false }
+          secret = nil
+
+          case algorithm
+          when Auth0::Algorithm::RS256
+            options[:jwks] = JSON.parse(JSON[algorithm.jwks], symbolize_names: true)
+          when Auth0::Algorithm::HS256
+            secret = algorithm.secret
+          end
```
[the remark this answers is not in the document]
Widcket: Addressed in efdf3feec312ed9d2e8a91fa3ea9b625373033ec

lib/auth0/mixins/validation.rb
```diff
@@ -20,6 +28,118 @@ def validate_permissions_array(permissions)
...
+          validate_signature decoded_id_token
+          validate_claims decoded_id_token
+        end
+
+        private
+
+        def validate_signature(decoded_id_token)
+          algorithm = @context[:algorithm]
+          algorithm_name = algorithm.class.to_s.split('::').last
+          options = { algorithms: [algorithm_name], verify_expiration: false, verify_not_before: false }
```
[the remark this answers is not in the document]
Widcket: Added in efdf3feec312ed9d2e8a91fa3ea9b625373033ec

lib/auth0/mixins/validation.rb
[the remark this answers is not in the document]
Widcket: Changed in efdf3feec312ed9d2e8a91fa3ea9b625373033ec

lib/auth0/api/authentication_endpoints.rb
Widcket: Changed in efdf3feec312ed9d2e8a91fa3ea9b625373033ec

lib/auth0/mixins/validation.rb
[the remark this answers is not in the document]
Widcket: It's probably cleaner to leave it the way it is, since the decoding step checks for its own exceptions and conceptually decoding and signature validation are two distinct operations. The decoding is lightweight enough to not matter if it's being done twice.

Widcket:
> I left a few comments mostly centered around the cache implementation. In a cache miss scenario, it looks like a Key Rotation enabled tenant could be locked out for 10 minutes. But you might be able to confirm (and test) that.

Can you please further elaborate on this scenario?

lib/auth0/mixins/validation.rb
```diff
@@ -66,20 +72,27 @@ def validate_signature(decoded_id_token)
...
         end
+        # rubocop:enable Metrics/MethodLength, Metrics/AbcSize
 
-        def validate_claims(decoded_id_token); end
+        def validate_claims(decoded_id_token); end # TODO: Implement (in the next PR)
+      end
+
+      class JWTAlgorithm
+        def name
+          self.class.to_s.split('::').last
```
lbalmaceda, line 88: This is wrong. Use a string to define the name of the class in a class property.

lib/auth0/api/authentication_endpoints.rb
```diff
@@ -532,18 +532,14 @@ def validate_id_token(id_token, algorithm: nil, leeway: 60, nonce: nil, max_age:
...
-        unless algorithm.nil? || algorithm.is_a?(Auth0::Algorithm::RS256) || algorithm.is_a?(Auth0::Algorithm::HS256)
-          raise Auth0::InvalidParameter, "Signature algorithm of \"#{algorithm}\" is not supported"
-        end
-
         context = {
           issuer: issuer || "https://#{@domain}/",
           audience: audience || @client_id,
           algorithm: algorithm || Auth0::Algorithm::RS256.jwks_url("https://#{@domain}/.well-known/jwks.json"),
-          leeway: leeway
+          leeway: leeway,
```
lbalmaceda, line 539: We agreed you will change this back to how it was before.

lib/auth0/mixins/validation.rb
```diff
@@ -47,10 +47,16 @@ def validate(id_token)
 
         private
 
+        # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
         def validate_signature(decoded_id_token)
```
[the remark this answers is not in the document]
Widcket: Changed in 64bc3aa057870bb29267f94fd0333c74b958ed82

lib/auth0/api/authentication_endpoints.rb
Widcket: Changed in 64bc3aa057870bb29267f94fd0333c74b958ed82

lib/auth0/mixins/validation.rb
Widcket: Changed in 64bc3aa057870bb29267f94fd0333c74b958ed82

lib/auth0/mixins/validation.rb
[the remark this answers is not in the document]
Widcket: Changed in 64bc3aa057870bb29267f94fd0333c74b958ed82

PR #223, opened 2020-06-08
Improve OIDC compliance: add claims validation support [SDK-987]

Widcket:
### Changes

This update improves the SDK support for **OpenID Connect (OIDC)**. In particular, it adds support for **claims validation**.

#### What’s being added in this PR

- The implementation of the `validate_claims` method. Plus tests.

#### What’s being changed in this PR

- Moved the params validation to the validator class, along with its tests.
- Replaced the public API test cases with ones that test just the logic inside that method.

### Testing

### Checklist

### DIFF
--- a/lib/auth0/mixins/validation.rb
+++ b/lib/auth0/mixins/validation.rb
@@ -149,7 +149,7 @@ module Auth0
                                          " found \"#{claims['aud']}\""
           elsif claims['aud'].is_a?(Array) && !claims['aud'].include?(expected)
             raise Auth0::InvalidIdToken, "Audience (aud) claim mismatch in the ID token; expected \"#{expected}\""\
-                                         " but was not one of \"#{claims['aud']}\""
+                                         " but was not one of \"#{claims['aud'].join ', '}\""
           end
         end
 
--- a/lib/auth0/mixins/validation.rb
+++ b/lib/auth0/mixins/validation.rb
@@ -158,7 +158,7 @@ module Auth0
             raise Auth0::InvalidIdToken, 'Expiration Time (exp) claim must be a number present in the ID token'
           end
 
-          now = @context[:clock] || Time.now
+          now = @context[:clock] || Time.now.to_i
           exp_time = claims['exp'] + leeway
 
           unless now < exp_time
@@ -172,7 +172,7 @@ module Auth0
             raise Auth0::InvalidIdToken, 'Issued At (iat) claim must be a number present in the ID token'
           end
 
-          now = @context[:clock] || Time.now
+          now = @context[:clock] || Time.now.to_i
           iat_time = claims['iat'] - leeway
 
           unless now > iat_time
@@ -209,7 +209,7 @@ module Auth0
                                          ' token when Max Age (max_age) is specified'
           end
 
-          now = @context[:clock] || Time.now
+          now = @context[:clock] || Time.now.to_i
           auth_valid_until = claims['auth_time'] + max_age + leeway
 
           unless now < auth_valid_until
--- a/lib/auth0/mixins/validation.rb
+++ b/lib/auth0/mixins/validation.rb
@@ -28,6 +28,7 @@ module Auth0
         permissions.map { |permission| permission.to_h }
       end
 
+      # rubocop:disable Metrics/ClassLength
       class IdTokenValidator
         def initialize(context)
           @context = context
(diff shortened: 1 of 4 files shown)


lib/auth0/mixins/validation.rb
```diff
@@ -85,15 +86,140 @@ def decode_and_validate_signature(id_token, header)
...
+          issuer = @context[:issuer]
+          audience = @context[:audience]
+          max_age = @context[:max_age]
+
+          raise Auth0::InvalidParameter, 'Must supply a valid leeway' unless leeway.is_a?(Integer) && leeway >= 0
+          raise Auth0::InvalidParameter, 'Must supply a valid nonce' unless nonce.nil? || !nonce.to_s.empty?
+          raise Auth0::InvalidParameter, 'Must supply a valid issuer' unless issuer.nil? || !issuer.to_s.empty?
+          raise Auth0::InvalidParameter, 'Must supply a valid audience' unless audience.nil? || !audience.to_s.empty?
+
+          unless max_age.nil? || (max_age.is_a?(Integer) && max_age >= 0)
```
lbalmaceda, line 109: this one looks similar to the first check a few lines above for leeway. What's the reason for using a different style here? remember leeway could also be set to `nil` by the dev

lib/auth0/mixins/validation.rb
```diff
@@ -85,15 +86,140 @@ def decode_and_validate_signature(id_token, header)
...
+            raise Auth0::InvalidParameter, 'Must supply a valid max_age'
+          end
+
+          validate_iss(claims, issuer)
```
lbalmaceda, line 113: these validation methods don't need the entire claims map. You could reduce the scope if you pass the value and check for presence comparing against nil, for example.

spec/lib/auth0/api/authentication_endpoints_spec.rb
```diff
@@ -678,55 +678,25 @@
...
 
-      expect { @instance.validate_id_token(token, algorithm: Auth0::Algorithm::RS256.jwks_url(JWKS_URL)) }.to_not raise_exception
       expect do
         @instance.validate_id_token(token,
-                                    algorithm: Auth0::Algorithm::RS256.jwks_url(JWKS_URL),
-                                    leeway: 10,
+                                    algorithm: Auth0::Algorithm::HS256.secret('secret'),
+                                    leeway: 100,
                                     nonce: 'nonce',
+                                    max_age: 2538301146,
```
lbalmaceda, line 696: if this value has a meaning, move it to a const (like you did for the mocked clock). Otherwise, I suggest using a lower number with less randomness as you do for the leeway.

lib/auth0/mixins/validation.rb
Widcket: It's because a single line would get too long.

lib/auth0/mixins/validation.rb
Widcket: It's done that way to ensure all the logic for verifying a particular claim is inside a single method.

spec/lib/auth0/api/authentication_endpoints_spec.rb
Widcket: The purpose of that particular test case is to test just the logic of the public API method, not the entire validation logic (that is already covered by the unit tests). That large value was deliberate and has no particular meaning, beyond getting the time-based validations to pass (because it's not possible to use a mocked clock with the public API).

lbalmaceda: It should be possible to pass a clock (for testing purposes) through the public API. All the SDKs do that and default to time.now()

lib/auth0/mixins/validation.rb
lbalmaceda: > It's done that way to ensure all the logic for verifying a particular claim is inside a single method.

Below the `azp` validation has a few conditions being checked right here

```ruby
validate_azp(claims, audience) if claims['aud'].is_a?(Array) && claims['aud'].count > 1
```

If you continue passing the `claims` to the `validate_azp` method, you could check those conditions inside the method instead, to _really_ keep all the logic for verifying a claim in a single method.

The alternative I was suggesting was passing the value instead of the map.

```ruby
validate_azp(claims['azp'], audience) if claims['aud'].is_a?(Array) && claims['aud'].count > 1
```

As it is today, it is a mix of both.

spec/lib/auth0/api/authentication_endpoints_spec.rb
Widcket: In this case it doesn't make sense IMO, since all the time-based testing is actually done with a mocked clock. in this particular test case the claims validation is not what's being tested, and adding an extra parameter to the public API just for this test case it's a bit too much, I think.

lib/auth0/mixins/validation.rb
Widcket: The leeway is one param that is not expected to be `nil`, since it has a default value. Of course the validation will fail if it's `nil`, since it checks that the value is an integer, but is not a parameter in which `nil` is a valid option. If the dev does not want a leeway, the way to do that is to set it to `0`.

lib/auth0/mixins/validation.rb
Widcket: Those bits of conditional logic are checking preconditions. The claims validations logic is still all in the method.

PR #245, opened 2020-10-20
Remove deprecations

davidpatrick:
- Remove deprecated auth endpoints
- Remove v1 Endpoints
- Move RateLimitEncountered error class from Auth0::Unsupported to Auth0::Exception


lib/auth0/exception.rb
```diff
@@ -44,17 +44,12 @@ class MissingParameter < Auth0::Exception; end
...
   class InvalidCredentials < Auth0::Exception; end
   # Invalid Auth0 API namespace
   class InvalidApiNamespace < Auth0::Exception; end
   # Auth0 API rate-limiting encountered
-  # TODO: When making API-breaking changes, make this a subclass
-  # of Auth0::HTTPError directly rather than Auth0::Unsupported.
-  # It's currently under Unsupported to avoid breaking compatibility
-  # with prior gem versions that treated 429 errors as unknown errors.
-  class RateLimitEncountered < Auth0::Unsupported
+  class RateLimitEncountered < Auth0::Exception
```
lbalmaceda, line 52:
```
TODO: When making API-breaking changes, make this a subclass
  # of Auth0::HTTPError directly rather than Auth0::Unsupported.
```
Looks like the parent class should be different, have you check that?

davidpatrick: Yes you are correct, this should be `Auth0::HTTPError`

lbalmaceda (dismissed):

lbalmaceda (approved):

PR #247, opened 2020-10-21
Addressable dependency

davidpatrick:
### Changes
Addressable dependency was only avail in development environments through webmock.  It should be included in the Gemfile

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+## [v4.17.1](https://github.com/auth0/ruby-auth0/tree/v4.17.1) (2020-10-21)
+
+[Full Changelog](https://github.com/auth0/ruby-auth0/compare/v4.17.0...v4.17.1)
+
+**Fixed**
+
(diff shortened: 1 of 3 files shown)


stevehobbsdev (dismissed):

rally25rs:
Should `addressable` be added to the `gemspec` file too? IIRC `bundler` will refer to that, not the Gemfile, when adding `ruby-auth0` as a dependency to other consumer projects

rally25rs:
> Addressable dependency was only avail in development environments through webmock

That is funny. I was wondering how the CI passed, and so did ours after upgrading the gem. It only failed once we pushed it to a prod-like environment... but we use `webmock` too 😆

davidpatrick:
> That is funny. I was wondering how the CI passed, and so did ours after upgrading the gem. It only failed once we pushed it to a prod-like environment... but we use webmock too 😆

😅  thanks for catching it @rally25rs really appreciate it

rally25rs (dismissed):
Looks like it should work. Thanks for the super fast turnaround! 👍 Once a new gem version is published I'll give it a test in the same project & environment where we found the issue.

jimmyjames (approved):
LGTM

PR #197, opened 2019-09-25
Release 4.9.0

joshcanhelp:
See [milestone](https://github.com/auth0/ruby-auth0/issues?q=is%3Aclosed+milestone%3Av4.9.0) or CHANGELOG below for fixes and additions in this release. 

This PR also includes updates to development gems.

PR #254, opened 2020-12-24
relax dependency on rest-client. fixes #252

### LINKED ISSUE #252: Do not over-pin versions of dependencies
Is there a reason to pin versions up to the 3rd decimal?

```ruby
s.add_runtime_dependency 'rest-client', '~> 2.0.0'
s.add_runtime_dependency 'jwt', '~> 2.2.0'
s.add_runtime_dependency 'zache', '~> 0.12.0'
s.add_runtime_dependency 'addressable', '~> 2.7.0'
```

It prevents updating to `5.0.0` in my case because I had more up to date versions of those dependencies in my Gemfile (`rest-client` in my case). This results in `bundle update` updating to 4.8.0 only.

Pinning to the second `~> 2.0` would fix that.

Example:
```
Bundler could not find compatible versions for gem "rest-client":
  In snapshot (Gemfile.lock):
    rest-client (= 2.1.0)

  In Gemfile:
    rest-client

    auth0 (= 5.0.0) was resolved to 5.0.0, which depends on
      rest-client (~> 2.0.0)
```

lostapathy:
### Changes

Relaxes the dependency on rest-client, as requested in #252

### Checklist


davidpatrick (approved):

davidpatrick (changes requested):
Looks like we have some failing tests on this PR

lostapathy:
Looks like this was locked to the older version intentionally on #197.  I was hoping with the discussion on #252 that this would "just work" to relax it.

I've bisected it down to https://github.com/rest-client/rest-client/commit/197bb96c94ac90cbe8df7ab569bf8a2f2f501945  causing the issue, although I'm not yet sure how to fix it. Hopefully that rings a bell for someone more familiar with this gem.

lostapathy:
Related: https://github.com/rest-client/rest-client/issues/722

lostapathy:
I just pushed a (crude) workaround to this PR that fixes the tests.

I feel like there has to be a better way to fix this issue, hopefully this and the related issue above will help get it closer.

davidpatrick:
@lostapathy thank you for taking the time to raise this and doing a bit of research into it.  It looks like we actually pinned to the patch version because of this exact issue https://github.com/auth0/ruby-auth0/commit/f58d898510c31d6298a92b41603c61010254d88f I'm going to go ahead and close this PR and the associated issue until this issue is fixed in `rest-client`.

lostapathy:
@davidpatrick fair enough - but reading over on rest-client, I don't think they see this as a bug they need to fix in rest-client - they see it as a consumer issue that will need to be addressed in this library sooner or later.

Issue #189, opened 2019-08-21
Unable to set audience


rally25rs:
### Description

When initializing the client, the `audience` cannot be set.

If you try to use the default `Auth0 Management API` system api against a custom domain, you have to specify an `audience` that is different than the `domain`. You are currently unable to do this.

### Prerequisites

* [x] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
* [x] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md)
* [x] Did you check the documentation ([repo README](https://github.com/auth0/ruby-auth0/blob/master/README.md) or [Quickstart](https://auth0.com/docs/quickstart/backend/rails))?
* [x] Did you check the [Auth0 Community](https://community.auth0.com/)?
* [x] Are you reporting this to the correct repository? See also the [OmniAuth-Auth0 strategy](https://github.com/auth0/omniauth-auth0) for logging in with Rails.
* [x] Are there any related or duplicate [Issues](https://github.com/auth0/ruby-auth0/issues) or [PRs](https://github.com/auth0/ruby-auth0/pulls) for this issue?

### Environment

Please provide the following:

* Ruby Auth0 version: current `master` at the time of issue creation (`4.8.0`?)
* Ruby version:
* Rails version (if applicable):
* Browser version (if applicable):
* Additional gems that might be affecting your instance

### Reproduction

The code (which tries to pass an audience):
```
Auth0Client.new(
          client_id: '...',
          client_secret: '...',
          domain: 'accounts.my-org.com',
          audience: 'https://my-org.auth0.com/api/v2/',
          api_version: 2
```
results in the request body:
```
{
  "grant_type":"client_credentials",
  "client_id":"...",
  "client_secret":"...",
  "audience":"https://accounts.my-org.com/api/v2/"
}
```

The `audience` appears to be built based on the `domain` with no way to override it.

joshcanhelp:
@rally25rs - I'm guessing that you're using the built-in client credentials grant to get an API token for management API calls, is that correct? So initializing `Auth0::Client` without a manually-generated token?

If so, you have two options ... you can extend `Auth0::Api::AuthenticationEndpoints` and call the `api_token` method directly with in an `audience` there:

[...]

PR #261, opened 2021-03-07
Add api_identifier as an accepted configurable

QWYNG:
### Changes
I'd like to fetch a api token with audience that my api identifier.
When I use auth0 with custom domain, I need to specify an audience that is different than the domain. 
But I can not unable to do this now.

The `:audience` parameter is synonymous with the OIDC ID Token and is also used to generate the login URL, so it's confusing.
Auth0's Client Credentials Flow asks the audience for[ an API identifier](https://auth0.com/docs/api/authentication#client-credentials-flow).

`:custome_domein` and `:sub` were suggested at #189 , but I thought there was a more direct way to express it.
so I think the parameter `:api_identifier` is not confusing, What are your thoughts?

### References

This issue has been discussed at #189  as well.

### Testing

### Checklist

### DIFF
--- a/spec/lib/auth0/client_spec.rb
+++ b/spec/lib/auth0/client_spec.rb
@@ -29,7 +29,7 @@ describe Auth0::Client do
   let(:client_id) { '__test_client_id__' }
   let(:client_secret) { '__test_client_secret__' }
   let(:access_token) { '__test_access_token__' }
-  let(:audience) { "https://#{domain}/api/v2/" }
+  let(:organization) { '__test_organization__'}
 
   describe 'V2 client with token' do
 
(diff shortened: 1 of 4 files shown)


spec/lib/auth0/client_spec.rb
```diff
@@ -84,13 +83,12 @@
...
     end
 
-    context 'with token, audience, and client_secret' do
+    context 'with token and client_secret' do
       let(:subject) do
         Auth0::Client.new(
           token: access_token,
           domain: domain,
           client_secret: client_secret,
-          audience: audience
```
QWYNG, line 93: The audience parameter was used in the test, but as far as I investigated, this option has never been used for anything before.

davidpatrick (approved):

QWYNG:
@davidpatrick thank you!

nigelwtf:
Super late to the party on this one - but thought it better to comment than to raise a new issue.

Using a different audience & domain is probably a really common requirement - pretty much anyone using this library and who has a custom domain, but I only figured out this existed by reading through unit tests (I even tried Googling it afterwards just to make sure 😂).

Would it be worthwhile documenting `api_identifier` as a way to specify the `audience`? I'd be happy to raise a PR if someone steers me in the right direction.

PR #272, opened 2021-04-23
Properly handle headers when parsing x_ratelimit_reset in exception

benoittgt:
I noticed that we had type error when we were trying to get the `.reset`
time in an RateLimitEncountered. Headers are symbolized and are not
strings.

To reproduce de behavior I did
```ruby
t = []
  85.times do
    t << Thread.new do
      auth0_client.clients
    rescue Auth0::RateLimitEncountered => e
      puts <<~DOC

        exception: #{e.inspect}

        reset: #{e.reset}

      DOC
    end
  end
end; nil
p t.map(&:join); nil
p 1
```

Without the fix
```
.../gems/activesupport-6.0.3.6/lib/active_support/core_ext/time/calculations.rb:53:in `at': can't convert nil into an exact number (TypeError)
```

Inspecting headers:
```
headers['X-RateLimit-Reset']: nil, headers[:x_ratelimit_reset]: 1619192310
```

With the fix:
```
exception: #<Auth0::RateLimitEncountered: {"statusCode":429,"error":"Too Many Requests","message":"Global limit has been reached","errorCode":"too_many_requests"}>

reset: 2021-04-23 15:49:17 UTC
```

Related:
  - https://github.com/auth0/ruby-auth0/pull/170

---

### Checklist

### DIFF
--- a/lib/auth0/exception.rb
+++ b/lib/auth0/exception.rb
@@ -53,7 +53,7 @@ module Auth0
   # Auth0 API rate-limiting encountered
   class RateLimitEncountered < Auth0::HTTPError
     def reset
-      Time.at(headers['X-RateLimit-Reset']).utc
+      Time.at(Integer(headers[:x_ratelimit_reset])).utc
     end
   end
 
--- a/spec/lib/auth0/mixins/httpproxy_spec.rb
+++ b/spec/lib/auth0/mixins/httpproxy_spec.rb
@@ -104,9 +104,9 @@ describe Auth0::Mixins::HTTPProxy do
       it "should raise Auth0::RateLimitEncountered on send http #{http_method} method
         to path defined through HTTP when 429 recieved" do
         headers = {
-          'X-RateLimit-Limit'     => 10,
-          'X-RateLimit-Remaining' => 0,
-          'X-RateLimit-Reset'     => 1560564149
+          :x_ratelimit_limit     => 10,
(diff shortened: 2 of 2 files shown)


lib/auth0/exception.rb
```diff
@@ -53,7 +53,7 @@ class InvalidApiNamespace < Auth0::Exception; end
   # Auth0 API rate-limiting encountered
   class RateLimitEncountered < Auth0::HTTPError
     def reset
-      Time.at(headers['X-RateLimit-Reset']).utc
+      Time.at(Integer(headers[:x_ratelimit_reset])).utc
```
dmathieu, line 56: Any reason not to use `to_i` instead? It seems it would be more idiomatic.
It would also avoid raising an exception if the header is somehow empty.

benoittgt: Hello @dmathieu

Thanks for the review.

With `Integer` I wanted to avoid `''.to_i #=> 0` and then
```
Time.at(' '.to_i)
=> 1970-01-01 01:00:00 +0100
```

If we want to accept empty header value, I think we should early return
```ruby
x_ratelimit_reset = headers[:x_ratelimit_reset]
return if x_ratelimit_reset.nil? || x_ratelimit_reset.empty?

Time.at(headers[:x_ratelimit_reset]).utc
```

davidpatrick (approved):

Issue #277, opened 2021-07-13
The dependency on addressable '~> 2.7.0' is forcing the usage of an insecure gem


cyclotron3k:
The latest version of `ruby-auth0` has the following line in `auth0.gemspec`:

    s.add_runtime_dependency 'addressable', '~> 2.7.0'

Version 2.7.0 of `addressable` has recently had a CVE reported: https://github.com/advisories/GHSA-jxhc-q857-3j6g
And this version constraint makes it impossible to upgrade to the secure version (2.8.0)

Suggest using a more relaxed version constraint:

    s.add_runtime_dependency 'addressable', '~> 2.7'

Widcket:
The `v5.1.2` release [is out](https://rubygems.org/gems/auth0/versions/5.1.2) now, which uses addressable `2.8.0`.

PR #279, opened 2021-07-13
Do not lock dependencies to minor version

athix:
Overly restrictive version locks will [cause unnecessary issues for downstream applications](https://github.com/auth0/ruby-auth0/issues/277). All gems should be following semver, such that minor version bumps are completely backwards compatible. By fuzzy locking to the minor version that you need, you both allow upgrading to compatible newer versions, while also ensuring the inclusion of functionality that may have been introduced in a minor version.

### DIFF
--- a/auth0.gemspec
+++ b/auth0.gemspec
@@ -16,7 +16,7 @@ Gem::Specification.new do |s|
   s.executables   = `git ls-files -- bin/*`.split("\n").map { |f| File.basename(f) }
   s.require_paths = ['lib']
 
-  s.add_runtime_dependency 'rest-client', '~> 2.0'
+  s.add_runtime_dependency 'rest-client', '~> 2.0.2' #2.1.0 has breaking changes
   s.add_runtime_dependency 'jwt', '~> 2.2'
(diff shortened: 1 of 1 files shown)


athix:
It appears I can't see why CircleCI is failing without creating an account. Github actions avoids this issue, and provides pretty in-depth CI/CD integrations.

cyclotron3k:
Would be good to take another look at this one. The overly restrictive version locking caused a issue as predicted: https://github.com/auth0/ruby-auth0/issues/277

davidpatrick:
Hey @athix thanks for raising the PR.  The tests are failing locally as well.  It looks to be an issue with rest client 2.1.0

davidpatrick (approved):

athix:
> Hey @athix thanks for raising the PR. The tests are failing locally as well. It looks to be an issue with rest client 2.1.0

:man_facepalming: It's frustrating that rest-client isn't following semver... Also I took a look and it seems like the library might be unmaintained at the moment, which is a bit concerning. I created an issue to ask about this: https://github.com/rest-client/rest-client/issues/764

Thanks for tracking down what the problem was! :slightly_smiling_face:

PR #287, opened 2021-07-27
Change strategy when normalizing path

davidpatrick:
When encoding the uri, we currently have two different strategies, either the full URL or the path.  When passing in just the path, the Addressable aggresively parses the path assuming its a full url.  This causes some charachters to not be encoded and instead be parsing points of the url.

This PR changes the strategy to check for when the base_uri is set, and if is is we dont need to parse the url, just encode the path.

### DIFF
--- a/lib/auth0/mixins/httpproxy.rb
+++ b/lib/auth0/mixins/httpproxy.rb
@@ -9,8 +9,15 @@ module Auth0
 
       # proxying requests from instance methods to HTTP class methods
       %i(get post post_file put patch delete delete_with_body).each do |method|
-        define_method(method) do |path, body = {}, extra_headers = {}|
-          safe_path = Addressable::URI.escape(path)
+        define_method(method) do |uri, body = {}, extra_headers = {}|
+
+          if base_uri
+            # if a base_uri is set then the uri can be encoded as a path
+            safe_path = Addressable::URI.new(path: uri).normalized_path
+          else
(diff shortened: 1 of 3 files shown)


stevehobbsdev (commented):
Are there any tests left that just test the path encoding or have they all been changed to use a full URL?

davidpatrick:
The tests were previously incorrectly written .  As you can see on https://github.com/auth0/ruby-auth0/pull/287/files#diff-1b8f5b737fbf10530ab69022fd4c899103d53ab4dc81c45d52aad02b9f4a3ef4L4 it was attempting to set the base_uri path, but incorrectly doing so.  This resulted in the test expectations being changed to not include the base uri.  After correctly setting the base uri, all these tests start failing, and there for need to include the full base_uri in the test.

The change to the code here isnt whether or not the request is made to a full URL, its whether or not HttpProxy is being passed a full url or a path.  These tests are *still* passing in only a path `@instance.send(http_method, '/te st#test')`, just the result is the base uri followed with the correctly encoded path (before it would have not correctly encoded to `https://auth0.com/te%20st%23test` https://github.com/auth0/ruby-auth0/pull/287/files#diff-d46c5ee08b30cd5f46f76798c16bae372520050bb9263e6b4225f8bc84aa3956R148)

stevehobbsdev (approved):

PR #290, opened 2021-08-05
[SDK-2689] Automatically retry requests when API returns a 429 rate-limit status header. 

davidpatrick:
### Changes

This PR enables support for automatic retry of API requests that encounter a 429 rate-limit status response from the Auth0 API. It will retry, by default, up to 3 times, with a hard-coded cap of 10. It exponentially increases the delay between each request ("exponential backoff"), and uses a jitter technique to slightly randomize the variance of this delay.

### References

Please see the internal SDK-2653 epic on this topic.

### Testing

Coverage added ✅ 
existing tests also ✅


evansims (approved):
Awesome work! LGTM 👍

PR #297, opened 2021-09-08
Deprecate change_password in favor of reset_password

### LINKED ISSUE #292: Unsupported parameter `password` for `change_password` API
### Describe the problem
In trying to implement a reset password workflow in our application, I found the `/dbconnections/change_password` endpoint which this gem implements in the `change_password` method.

The [auth0 documentation](https://auth0.com/docs/api/authentication#change-password) for this endpoint lists 3 parameters:
- `client_id`
-  `email`
-  `connection`

The documentation in the gem itself lists `password` as a possible parameter. This parameter is not listed in the API documentation.

https://github.com/auth0/ruby-auth0/blob/2a5cb4cba7c26b5e1bb7e6767fa5a31e28e80354/lib/auth0/api/authentication_endpoints.rb#L149-L165

Attempting to call `change_password` with a value for the `password` parameter results in a `Auth0::BadRequest` error, at least for me:

```
/Users/zach/.gem/ruby/3.0.2/gems/auth0-5.5.0/lib/auth0/mixins/httpproxy.rb:93:in `request': {"error":"password is not allowed"} (Auth0::BadRequest)
```

### What was the expected behavior?
I think this behavior is probably expected considering the gem implementation doesn't match what the API documentation lists as the interface for this endpoint.

### Reproduction
1. Create an Auth0 user
2. Attempt to change that user's email, following the `change_password` method documentation:
```ruby
auth0_client.change_password('youremail@here.com', 'thisIsMyNewP4ssw0rd') 
```
3. See `Auth0::BadRequest` returned

### Environment

- **Version of this library used:**
5.5.0
- **Which framework are you using, if applicable:**
Rails
- **Other modules/plugins/libraries that might be involved:**
N/A
- **Any other relevant information you think would be useful:**
Ruby 3.0.2

davidpatrick:
The method `change_password` took a password argument that was only supported on legacy tenants with the flag `change_pwd_flow_v1` enabled.  In order to resolve confusion, we're introducing a new method `reset_password` to fire off a password reset email, and deprecating the old method.

Fixes #292

### DIFF
--- a/lib/auth0/api/authentication_endpoints.rb
+++ b/lib/auth0/api/authentication_endpoints.rb
@@ -151,7 +151,7 @@ module Auth0
       # @see https://auth0.com/docs/connections/database/password-change
       # @param email [string] User's current email
       # @param password [string] User's new password. This is only available
-      #   on legacy tenants with the change_pwd_flow_v1 flag enabled
+      #   on legacy tenants with change password v1 flow enabled
       # @param connection_name [string] Database connection name
       # @deprecated Use {#password_reset} instead.
       def change_password(email, password, connection_name = UP_AUTH)
--- a/lib/auth0/api/authentication_endpoints.rb
+++ b/lib/auth0/api/authentication_endpoints.rb
@@ -150,15 +150,34 @@ module Auth0
       # @see https://auth0.com/docs/api/authentication#change-password
       # @see https://auth0.com/docs/connections/database/password-change
       # @param email [string] User's current email
-      # @param password [string] User's new password; empty to trigger a
-      #   password reset email
+      # @param password [string] User's new password. This is only available
+      #   on legacy tenants with the change_pwd_flow_v1 flag enabled
       # @param connection_name [string] Database connection name
+      # @deprecated Use {#password_reset} instead.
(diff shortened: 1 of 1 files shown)


lib/auth0/api/authentication_endpoints.rb
```diff
@@ -150,9 +150,10 @@ def signup(email, password, connection_name = UP_AUTH)
       # @see https://auth0.com/docs/api/authentication#change-password
       # @see https://auth0.com/docs/connections/database/password-change
       # @param email [string] User's current email
-      # @param password [string] User's new password; empty to trigger a
-      #   password reset email
+      # @param password [string] User's new password. This is only available
+      #   on legacy tenants with the change_pwd_flow_v1 flag enabled
```
evansims, line 154: Honestly not sure, but do we expose the flag identifier anywhere in a customer-facing fashion? If not, it might be better to identify this as the "Change Password Flow" toggle indicated in their tenant account settings, under the advanced tab: if v2 is enabled it's locked out for them.

evansims:
Question aside LGTM

evansims (dismissed):

evansims (approved):

PR #648, opened 2025-09-17
Make Auth0::Client#get_token public

ttstarck:
#### Code Changes
Make `Auth0::Client#get_token` public.

#### Why
Currently if I want to generate an API token for an API using the `Auth0::Client`, I must do the following:
```ruby
 # This makes a request to `/oauth/token`
client = Auth0::Client.new(api_identifier: "my-api-identifier", client_id: "my-client-id", client_secret: "my-client-secret", domain: "my-domain.auth0.com")

# This makes a duplicate request to `/oauth/token`
api_token = client.api_token("my-api-identifier")

# i.e. make an http request to this API with the api token as the Authorization bearer.
do_something(api_token)
```

However `Auth0::Client.new` will actually already request this api token on initialize, however this token is not available via any public interface. It is only stored as an instance variable in `@token`.

This commit makes `Auth0::Client#get_token` public (formerly private).

Now when in need of an API token for the API, we can do the following:
```ruby
# This makes a request to `/oauth/token`
client = Auth0::Client.new(api_identifier: "my-api-identifier", client_id: "my-client-id", client_secret: "my-client-secret", domain: "my-domain.auth0.com")

 # This does not make another request to `/oauth/token` unless the existing token has expired.
api_token = client.get_token

# i.e make an http request to this API using the API token as the Authorization bearer.
do_something(api_token)
```

#### Why not make `@token` an `attr_reader`?

`#get_token` handles generating a new api token if the existing one has expired, saving me from having to handle that logic myself.

### References
None
### Testing

There already are tests for this method `#get_token`, I have just changed the tests to no longer use `send` as the method is public now.

* [ ] This change adds unit test coverage
* [ ] This change adds integration test coverage
* [x] This change has been tested on the latest version of Ruby

### Checklist

* [x] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
* [x] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md)
* [ ] All existing and new tests complete without errors
* [ ] Rubocop passes on all added/modified files
* [ ] All active GitHub checks have passed

PR #725, opened 2026-05-04
Make Auth0::Client#get_token public

arpit-jn:
## Summary
- Makes `Auth0::Client#get_token` a public method (previously private)
- Allows callers to retrieve the current API token without a redundant `/oauth/token` request
- Updates tests to call `get_token` directly instead of via `send(:get_token)`

This takes forward the community contribution from #648 by @ttstarck.

## References
- Original PR: #648

### DIFF
--- a/lib/auth0/mixins/token_management.rb
+++ b/lib/auth0/mixins/token_management.rb
@@ -1,22 +1,14 @@
 module Auth0
   module Mixins
     module TokenManagement
-      
-      private
 
-      def initialize_token(options)
-        @token = options[:access_token] || options[:token]
-        # default expiry to an hour if a token was given but no expires_at
-        @token_expires_at = @token ? options[:token_expires_at] || Time.now.to_i + 3600 : nil 
-
-        @audience = options[:api_identifier] || "https://#{@domain}/api/v2/"
-        get_token() if @token.nil?
-      end
-
+      # Get the Client's api token (or generate a new one if it has expired).
+      #
+      # @return [String] the api token
       def get_token
         # pp @token_expires_at
         has_expired = @token && @token_expires_at ? @token_expires_at < (Time.now.to_i + 10) : false
-        
+
         if (@token.nil? || has_expired) && @client_id && (@client_secret || @client_assertion_signing_key)
           response = api_token(audience: @audience)
           @token = response.token
@@ -26,6 +18,17 @@ module Auth0
         else
           @token
         end
+      end
+
+      private
+
+      def initialize_token(options)
+        @token = options[:access_token] || options[:token]
+        # default expiry to an hour if a token was given but no expires_at
+        @token_expires_at = @token ? options[:token_expires_at] || Time.now.to_i + 3600 : nil
+
+        @audience = options[:api_identifier] || "https://#{@domain}/api/v2/"
+        get_token() if @token.nil?
       end
     end
   end
--- a/lib/auth0/mixins/token_management.rb
+++ b/lib/auth0/mixins/token_management.rb
@@ -4,9 +4,9 @@ module Auth0
 
       # Get the Client's api token (or generate a new one if it has expired).
       #
+      # @note This method may perform a network request to refresh an expired token. It is not thread-safe.
       # @return [String] the api token
       def get_token
-        # pp @token_expires_at
         has_expired = @token && @token_expires_at ? @token_expires_at < (Time.now.to_i + 10) : false
 
         if (@token.nil? || has_expired) && @client_id && (@client_secret || @client_assertion_signing_key)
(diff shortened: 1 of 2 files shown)


lib/auth0/mixins/token_management.rb
```diff
@@ -1,22 +1,14 @@
...
-      def initialize_token(options)
-        @token = options[:access_token] || options[:token]
-        # default expiry to an hour if a token was given but no expires_at
-        @token_expires_at = @token ? options[:token_expires_at] || Time.now.to_i + 3600 : nil 
-
-        @audience = options[:api_identifier] || "https://#{@domain}/api/v2/"
-        get_token() if @token.nil?
-      end
 
+      # Get the Client's api token (or generate a new one if it has expired).
```
kishore7snehil, line 5: 🔵 **LOW** · conventions

The YARD doc says `@return [String] the api token` but does not document that the method may perform an HTTP request to `/oauth/token` and mutate internal state (`@token`, `@token_expires_at`). Since this is now a public API method, callers should know that invoking `get_token` may trigger network activity.

> Update the YARD doc to: `# Get the Client's api token (or generate a new one if it has expired). Note: this method may perform a network request to refresh an expired token and is not thread-safe.`

lib/auth0/mixins/token_management.rb
```diff
@@ -1,22 +1,14 @@
...
+      #
+      # @return [String] the api token
```
kishore7snehil, line 8: 🔵 **LOW** · conventions

The `@return` tag says `[String]` but the method can return `nil` when `@token` is nil and no `@client_id`/`@client_secret` is configured (the `else` branch returns `@token` which could be nil).

> Update to `@return [String, nil] the API token, or nil if no token is available and credentials are not configured`.

lib/auth0/mixins/token_management.rb
```diff
@@ -1,22 +1,14 @@
...
       def get_token
         # pp @token_expires_at
```
kishore7snehil: ℹ️ **INFO** · conventions

Leftover debug comment `# pp @token_expires_at` — this was pre-existing but now that the method is part of the public API, it looks out of place.

> Remove this line.

lib/auth0/mixins/token_management.rb
```diff
@@ -1,22 +1,14 @@
...
-        @token_expires_at = @token ? options[:token_expires_at] || Time.now.to_i + 3600 : nil 
-
-        @audience = options[:api_identifier] || "https://#{@domain}/api/v2/"
-        get_token() if @token.nil?
-      end
 
+      # Get the Client's api token (or generate a new one if it has expired).
+      #
+      # @return [String] the api token
       def get_token
```
kishore7snehil, line 9: 🔵 **LOW** · security

Making `get_token` public exposes a method that can trigger a network call and mutate state without synchronization. In multi-threaded environments (Puma, Sidekiq), concurrent callers could both observe `has_expired == true` and both issue `/oauth/token` requests (TOCTOU race). The worst outcome is a redundant HTTP request, not a credential leak — but this is worth documenting since the method is now public.

> Consider adding `# @note This method is not thread-safe. Concurrent callers may trigger redundant token refreshes.`

spec/lib/auth0/mixins/token_management_spec.rb
```diff
@@ -130,7 +130,7 @@
 
       expect(RestClient::Request).not_to receive(:execute)
```
kishore7snehil, line 132: ℹ️ **INFO** · testing

This test calls `instance.get_token` but only asserts that no HTTP request is made — it does not assert the return value. If `get_token` returned `nil` instead of the existing token, this test would still pass.

> Consider adding: `expect(instance.get_token).to eq('existing-token')`

kishore7snehil (changes requested):

kishore7snehil (approved):
LGTM!

lib/auth0/mixins/token_management.rb
arpit-jn: Addressed.

lib/auth0/mixins/token_management.rb
arpit-jn: This one could mislead. It tells callers to nil-guard against a scenario that means the client is broken. A misconfigured client should fail loudly at init, not silently return nil. Documenting nil as valid makes a bug look like expected behavior.

lib/auth0/mixins/token_management.rb
arpit-jn: Updated.

lib/auth0/mixins/token_management.rb
arpit-jn: Updated.

spec/lib/auth0/mixins/token_management_spec.rb
arpit-jn: Updated.

PR #766, opened 2026-07-15
fix: let Fern own client.rb so sub-client accessors stay in sync

kishore7snehil:
### Changes

Removes `lib/auth0/client.rb` (the root `Auth0::Management` client) from `.fernignore` so Fern owns it on regeneration.


developerkunal (approved):
LGTM

PR #774, opened 2026-07-27
Expose Management API rate limit info via a client callback

### LINKED ISSUE #606: Rate limit data returned from API methods
### Checklist

- [X] I have looked into the [Readme](https://github.com/auth0/ruby-auth0#readme) and [Examples](https://github.com/auth0/ruby-auth0/blob/master/EXAMPLES.md), and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://www.rubydoc.info/gems/auth0) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/ruby-auth0/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

We're integrating with Auth0 and we would like to gather rate limit data from our calls to the Authentication and Management APIs. We have monitors for when we are rate limited, but we would like to gather metrics on the rate limit data returned from each API call. This would give us clear data points on how close we are to the rate limits as we continue to roll it out to more of our users.

### Describe the ideal solution

I've partially implemented this as a proof of concept and I'd be happy to create a pull request for this if you're open to it:

- An optional block can be given to capture the response and rate limit
- The original method can still be called as is making this a non-breaking change

For example, both of these are valid method calls:

```
# Original method still works
auth0_client.user_sessions(auth0_id)
```

```
# We could use the returned `rate_limit_info` to create Datadog metrics
auth0_client.user_sessions(auth0_id) do |response_body, rate_limit_info|
  puts "Response body: #{response_body}"
  puts "Rate limit info: #{rate_limit_info}"
end
```

We're considering monkey patching something like this to gather the rate limit information, but we'd much rather contribute this change if you're open to it!

### Alternatives and current workarounds

_No response_

### Additional context

_No response_

ProdigyTom:
### Changes

Exposes Auth0's rate-limit information (`x-ratelimit-limit` / `-remaining` / `-reset`) from **every Management API response**, so callers can monitor how close they are to the limit (the ask in #606). It's delivered via an opt-in callback rather than a changed return value — see the note below on why.

- **`Auth0::Internal::Http::RateLimit`** — value object (`limit`, `remaining` as Integers; `reset` as a UTC `Time`). Blank/non-numeric header values become `nil` (never a misleading `0`).
- **`RawClient`** gains an optional `rate_limit_handler`, invoked in `#send` after retries on every response with the parsed `RateLimit`. Handler errors are swallowed so a monitoring callback can never break a request. `#send`'s return value is unchanged, so the generated callers are unaffected.
- **Wiring:** `Auth0::Client.new(..., management_rate_limit_handler: ->(rl) { ... })` attaches the handler to the management raw client.
- All changes live in fernignored files (`lib/auth0/internal/**`, `lib/auth0/mixins/**`, `lib/auth0/auth_client.rb`), so they survive regeneration.
- Unit tests for `RateLimit`, `RawClient#send` handler behavior, and the client wiring.

```ruby
client = Auth0::Client.new(
  domain: ENV["AUTH0_DOMAIN"],
  token: ENV["AUTH0_MGMT_TOKEN"],
  management_rate_limit_handler: lambda do |rate_limit|
    StatsD.gauge("auth0.mgmt.remaining", rate_limit.remaining) if rate_limit.remaining
  end
)
client.users.get(id: "auth0|123") # handler fires with the response's rate limit
```

**Why a callback instead of returning the rate limit (e.g. `with_raw_response`)?** The generated endpoint methods (`Auth0::Users::Client#get`, etc.) parse the body and discard the response, and they're regenerated — so a third-party change to what they *return* wouldn't survive. A callback invoked at the maintained `RawClient` layer is the part we can change durably while still delivering the data on every request. If you'd prefer to also expose it as a return value via a generated `with_raw_response`-style accessor, this `RateLimit` object slots right in — happy to align.

### References

- Refs #606 (the v5 implementation, #774, is superseded by the v6 rewrite)

### Testing

Unit tests under `test/unit/` cover header parsing (`0` vs blank/non-numeric), the handler firing with the parsed rate limit while the response is returned unchanged, handler errors not breaking the request, and the client wiring. Full `rake test` passes locally with no failures.

### Checklist

* [x] I have read the Auth0 general contribution guidelines
* [x] I have read the Auth0 Code of Conduct


kishore7snehil:
@ProdigyTom I can see some merge conflicts. Can you please resolve them? Post that I can take a look

ProdigyTom:
@kishore7snehil my original change was against v5 and after the change to v6 I wasn't really able to resolve my merge conflicts in a way that still solved our problem. The rate limiting information gets dropped earlier in some generated code. So I wound up going a slightly different direction. let me know if it makes sense

